PR #3, opened 2017-10-06
match deployment blocks from augur-contracts PR

epheph:
https://github.com/AugurProject/augur-contracts/pull/4


epheph:
@tinybike familiar with this CI failure?

MicahZoltu:
@epheph Your CI script needs to run `tsc` in the root of the project before running the tests.

epheph:
Oh, it looks like build was committed previously, that's why this just popped up.

@pgebheim

tinybike:
Shouldn't we retain the old upload block numbers, if we're also picking up LegacyRepContract events?

epheph:
yea, makes sense

PR #12, opened 2017-10-14
Feature/3862 heroku

pgebheim:
This includes the merges for #11 as well and some bug fixes -- so I am going to close that.


src/server/run-websocket-server.ts
```diff
@@ -1,12 +1,27 @@
...
-  const websocketServer: WebSocket.Server = new WebSocket.Server({ port });
+let app: any = express();
+let server: http.Server  = http.createServer(app);
+
+
+app.get('/', (req: any, res: any): void => {
+  res.send("Hello");
+});
+
+export function runHttpServer(port: number, callback: ErrorCallback): void {
```
tinybike: Why are we setting up an HTTP server?  MVP augur-node is intended to be websockets-only.  Is there something we need an HTTP server for?

tinybike: Went ahead and removed the HTTP server stuff so this can be merged -- if heroku needs this for some reason we can make it a separate PR.

PR #16, opened 2017-10-17
Generate declaration files

pgebheim:
Generate declaration files to start packaging this as a module

### DIFF
--- a/src/server/getters/get-account-transfer-history.ts
+++ b/src/server/getters/get-account-transfer-history.ts
@@ -1,7 +1,7 @@
 import * as Knex from "knex";
 import { Address, Bytes32 } from "../../types";
 
-export interface TransferRow {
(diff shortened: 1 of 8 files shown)


src/server/getters/get-account-transfer-history.ts
```diff
@@ -1,7 +1,7 @@
 import * as Knex from "knex";
 import { Address, Bytes32 } from "../../types";
 
-interface TransferRow {
+export interface TransferRow {
```
epheph: These seem internally used only, is there a reason to `export` them?

pgebheim: Correct, `TransferRow` is internal and I won't export it. `TransferHistory` and `TransferLog` are in the public interface.

PR #17, opened 2017-10-18
designated-report-submitted log processor

epheph:
Need to write tests, but I wanted a review of my work first, since it's my first log-processor commit, as well as the first primary/autoincrement table.

### DIFF
--- a/src/blockchain/log-processors/designated-report-submitted.ts
+++ b/src/blockchain/log-processors/designated-report-submitted.ts
@@ -19,16 +19,16 @@ export function processDesignatedReportSubmittedLog(db: Knex, augur: Augur, trx:
         blockNumber: log.blockNumber
     };
 
-    db.transacting(trx).insert(marketStateDataToInsert).returning("marketStateID").into("market_state").asCallback((err: Error|null, marketStateID: Array<number>): void => {
+    db.transacting(trx).insert(marketStateDataToInsert).returning("marketStateID").into("market_state").asCallback((err: Error|null, marketStateID?: Array<number>): void => {
         if (err) return callback(err);
-        if (!marketStateID) return callback(new Error("Failed to generate new marketStateID for marketID:" + log.market));
+        if (!marketStateID || !marketStateID.length) return callback(new Error("Failed to generate new marketStateID for marketID:" + log.market));
         const newMarketStateID = marketStateID[0];
         db("markets").transacting(trx).update({ marketStateID: newMarketStateID }).where("marketID", log.market).asCallback((err: Error|null): void => {
             if (err) return callback(err);
             const dataToInsert: { [index: string]: string|number } = {
-                marketID: log.market;
+                marketID: log.market
             };
-            log.payoutNumerators.forEach((value: number, i: number) => {
+            log.payoutNumerators.forEach((value: number, i: number): void => {
                 dataToInsert["payout" + i] = value;
             });
             db.transacting(trx).insert(dataToInsert).into("reports_designated").asCallback(callback);
--- a/src/blockchain/log-processors/designated-report-submitted.ts
+++ b/src/blockchain/log-processors/designated-report-submitted.ts
@@ -4,14 +4,12 @@ import { FormattedLog, ErrorCallback } from "../../types";
 
 export function processDesignatedReportSubmittedLog(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
     // @achapman: "The designated reporter purchases Stake Tokens in order to do the Designated Report"
-    // Should designatedReportSubmitted include StakedToken address?
+    // TODO: Should designatedReportSubmitted include StakedToken address?
 
-    // for phase, sqlite lacks an ENUM, but the idea here is that:
     //  phase:null = no report submitted
     //  phase:0 = designated report submitted
     //  phase:1 = round1
     //  phase:2 = round2
-    // if that seems reasonable, i'll update PR to import the enum and use name below
     const marketStateDataToInsert: { [index: string]: string|number|boolean } = {
         marketID: log.market,
         phase: 0,
--- a/src/blockchain/log-processors/designated-report-submitted.ts
+++ b/src/blockchain/log-processors/designated-report-submitted.ts
@@ -2,8 +2,37 @@ import Augur = require("augur.js");
 import * as Knex from "knex";
 import { FormattedLog, ErrorCallback } from "../../types";
 
+
 export function processDesignatedReportSubmittedLog(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
-    console.log("TODO: DesignatedReportSubmitted");
-    console.log(log);
-    callback(null);
+    // @achapman: "The designated reporter purchases Stake Tokens in order to do the Designated Report"
+    // Should designatedReportSubmitted include StakedToken address?
+
+    // for phase, sqlite lacks an ENUM, but the idea here is that:
+    //  phase:null = no report submitted
+    //  phase:0 = designated report submitted
+    //  phase:1 = round1
+    //  phase:2 = round2
+    // if that seems reasonable, i'll update PR to import the enum and use name below
+    const marketStateDataToInsert: { [index: string]: string | number | boolean } = {
+        marketID: log.market,
+        phase: 0,
+        isDisputed: false,
+        blockNumber: log.blockNumber
+    }
+
+    db.transacting(trx).insert(marketStateDataToInsert).returning('marketStateID').into("market_state").asCallback((err: Error | null, marketStateID: number[]): void => {
+        if (err) return callback(err);
+        if (!marketStateID) return callback(new Error("Failed to generate new marketStateID for marketID:" + log.market));
+        const newMarketStateID = marketStateID[0];
+        db("markets").transacting(trx).update({ marketStateID: newMarketStateID }).where("marketID", log.market).asCallback((err: Error | null): void => {
+            if (err) return callback(err);
+            const dataToInsert: { [index: string]: string | number } = {
+                marketID: log.market
+            };
+            log.payoutNumerators.forEach(function (value: number, i: number) {
+                dataToInsert['payout' + i] = value;
+            });
+            db.transacting(trx).insert(dataToInsert).into("reports_designated").asCallback(callback);
+        })
+    })
 }
--- a/src/blockchain/log-processors/designated-report-submitted.ts
+++ b/src/blockchain/log-processors/designated-report-submitted.ts
@@ -12,20 +12,20 @@ export function processDesignatedReportSubmittedLog(db: Knex, augur: Augur, trx:
     //  phase:1 = round1
     //  phase:2 = round2
     // if that seems reasonable, i'll update PR to import the enum and use name below
-    const marketStateDataToInsert: { [index: string]: string | number | boolean } = {
+    const marketStateDataToInsert: { [index: string]: string|number|boolean } = {
         marketID: log.market,
         phase: 0,
         isDisputed: false,
         blockNumber: log.blockNumber
     };
 
-    db.transacting(trx).insert(marketStateDataToInsert).returning("marketStateID").into("market_state").asCallback((err: Error | null, marketStateID: Array<number>): void => {
+    db.transacting(trx).insert(marketStateDataToInsert).returning("marketStateID").into("market_state").asCallback((err: Error|null, marketStateID: Array<number>): void => {
         if (err) return callback(err);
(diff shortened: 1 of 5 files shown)


src/migrations/20171015155157_reports_designated.ts
```diff
@@ -0,0 +1,26 @@
...
+      payout1 numeric,
+      payout2 numeric,
+      payout3 numeric,
+      payout4 numeric,
+      payout5 numeric,
+      payout6 numeric,
+      payout7 numeric,
+      "isInvalid" integer
+    )`).then( (): void => {
+      knex.schema.table("reports_designated", (table: Knex.AlterTableBuilder): void => {
```
epheph: This method for adding an autoincrement primary key doesn't work, even though I cribbed it from elsewhere in the project (reports); ideas why?

epheph:
The logic here is this:

when a designated report is submitted:
1.) New row goes into market_state, documenting that the designated report has been submitted, and at what block number.
2.) That markets gets "marketStateID" updated to point to that row (current market state changed, that value will be updated to point to those rows every time its state change, but the rows themselves will hang around for market history queries)
3.) reports_designated records new row with the payout info (although it should probably be linked to the staked token? Not current a part of event?)

epheph:
Actually, I don't see any tests for any log-processors. Perhaps if we want to do start testing these, it'd be its own story.

src/migrations/20171015155157_reports_designated.ts
pgebheim: You just found a bug. SQLite can't add a PRIMARY KEY column after the table has been made.

Irritating. I'll massage this into working.

epheph: Blargh.

Hmmm. i wonder the best way to get it to work properly with pgsql and sqlite then, since pgsql does it significantly differently.

epheph: BTW i did this in a different migration:

https://github.com/AugurProject/augur-node/pull/17/files#diff-0e33934ad8e27a733628d374a0e880e0R7

tinybike:
We definitely do want to test the log processors, I just haven't gotten around to writing any tests yet. @epheph do you want to tackle this / open up a CH issue for it?

src/migrations/20171017150457_market_state.ts
```diff
@@ -0,0 +1,18 @@
+import * as Knex from "knex";
+
+// PRIMARY KEY AUTOINCREMENT pgsql compat issue? TODO: use .primary()
```
tinybike: Please use the knex functions as needed to address compatibility issues.  Actually, there appear to be enough subtle differences between sqlite/pgsql that it might be sensible for us to solely use the knex wrappers unless we're doing an abnormally complicated query / really need the raw sql input capability.

epheph: Did you see the above conversation about this? There's a bug in knex i'm working around; it is unable to add a key like this after table create.

epheph:
https://app.clubhouse.io/augur/story/4245/write-tests-for-log-processors

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
+    //  phase:2 = round2
+    // if that seems reasonable, i'll update PR to import the enum and use name below
+    const marketStateDataToInsert: { [index: string]: string | number | boolean } = {
+        marketID: log.market,
+        phase: 0,
+        isDisputed: false,
+        blockNumber: log.blockNumber
+    };
+
+    db.transacting(trx).insert(marketStateDataToInsert).returning("marketStateID").into("market_state").asCallback((err: Error | null, marketStateID: Array<number>): void => {
```
tinybike: Just double-checking b/c there aren't tests for this yet -- is `Array<number>` the right type for `marketStateID` here?  (It's not `Array(<{ marketStateID: number }>)` or something fancy like that?)

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
+        if (err) return callback(err);
+        if (!marketStateID) return callback(new Error("Failed to generate new marketStateID for marketID:" + log.market));
```
tinybike: Is it possible for marketStateID to be an empty array here?  If so, should change this conditional to `if (!marketStateID || !marketStateID.length)`.

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
+        const newMarketStateID = marketStateID[0];
+        db("markets").transacting(trx).update({ marketStateID: newMarketStateID }).where("marketID", log.market).asCallback((err: Error | null): void => {
+            if (err) return callback(err);
+            const dataToInsert: { [index: string]: string | number } = {
+                marketID: log.market;
+            };
+            log.payoutNumerators.forEach((value: number, i: number) => {
```
tinybike: Should have `: void` return type.

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
+    // @achapman: "The designated reporter purchases Stake Tokens in order to do the Designated Report"
+    // Should designatedReportSubmitted include StakedToken address?
+
+    // for phase, sqlite lacks an ENUM, but the idea here is that:
+    //  phase:null = no report submitted
+    //  phase:0 = designated report submitted
+    //  phase:1 = round1
+    //  phase:2 = round2
+    // if that seems reasonable, i'll update PR to import the enum and use name below
+    const marketStateDataToInsert: { [index: string]: string | number | boolean } = {
```
tinybike: Minor style point: elsewhere we've not put spaces around the `|` in types, i.e., `string|number` instead of `string | number`.  I don't have a strong preference for which style we use but I think it should be consistent throughout the code!

tinybike: (This occurs elsewhere too, but only explicitly mentioning it here.)

src/migrations/20171015155157_reports_designated.ts
```diff
@@ -0,0 +1,26 @@
+import * as Knex from "knex";
+
+exports.up = async (knex: Knex): Promise<any> => {
+  return knex.schema.dropTableIfExists("reports_designated").then( (): void  => {
```
tinybike: I believe @pgebheim's recent patch changed all names to use camel case names instead of snake case -- should probably use that convention here as well.  (Unless table names need to be exempt from that rule for some reason?)

src/blockchain/log-processors/designated-report-submitted.ts
epheph: This is the right type, i have tested this code with the event emitter contract

src/migrations/20171015155157_reports_designated.ts
tinybike: Also, minor style point: there is an extra space in `then( (): void  => {` between the two initial parens, and before the arrow.  As with other style gripes, I don't feel super strongly about this point, other than that I think the code base overall should have a uniform style.

tinybike: Note: style gripe occurs elsewhere as well, just bringing it up here.  Are there settings for this in tslint?  That would be ideal if so.

src/blockchain/log-processors/designated-report-submitted.ts
epheph: good call, my Python is showing...  in Python, `[]` is false.

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
+        if (err) return callback(err);
+        if (!marketStateID) return callback(new Error("Failed to generate new marketStateID for marketID:" + log.market));
+        const newMarketStateID = marketStateID[0];
+        db("markets").transacting(trx).update({ marketStateID: newMarketStateID }).where("marketID", log.market).asCallback((err: Error | null): void => {
+            if (err) return callback(err);
+            const dataToInsert: { [index: string]: string | number } = {
+                marketID: log.market;
+            };
+            log.payoutNumerators.forEach((value: number, i: number) => {
+                dataToInsert["payout" + i] = value;
```
tinybike: These column names are the gift that keeps on giving :grin:

src/blockchain/log-processors/designated-report-submitted.ts
tinybike: If the callback is passed an Error here, is `marketStateID` undefined / not passed in at all?  If so, the callback's signature should reflect that.

src/migrations/20171015155157_reports_designated.ts
epheph: table names are still under-score'd and lower (the only existing one i went off is `blockchain_sync_history`). Since they are represented as files on disk, not mixing case feels right, but i don't feel strongly. @pgebheim  ?

src/migrations/20171017150457_market_state.ts
tinybike: Ah!  Sorry about that, no I did not.  Is it possible to use knex to generate the create table command itself, and include the primary key autoincrement as part of that?

src/migrations/20171015155157_reports_designated.ts
epheph: re: extra space, you're right, i'm surprised linter didn't catch that? Anyway, i copied from:

https://github.com/epheph/augur-node/blob/d23310dfa7110c310e336e9d9972746682f9af67/src/migrations/20171011152617_reports.ts#L4

i'll fix there, too

src/blockchain/log-processors/designated-report-submitted.ts
```diff
@@ -3,7 +3,35 @@ import * as Knex from "knex";
...
-    callback(null);
+    // @achapman: "The designated reporter purchases Stake Tokens in order to do the Designated Report"
+    // Should designatedReportSubmitted include StakedToken address?
+
+    // for phase, sqlite lacks an ENUM, but the idea here is that:
+    //  phase:null = no report submitted
+    //  phase:0 = designated report submitted
+    //  phase:1 = round1
+    //  phase:2 = round2
+    // if that seems reasonable, i'll update PR to import the enum and use name below
```
tinybike: Hm, do you mean, import the enum in TS and convert (from phase number to phase name) in the application?  IMO null, 0, 1, and 2 are reasonably self-explanatory in this context so it might not be necessary to do anything fancy!

src/migrations/20171015155157_reports_designated.ts
tinybike: Hm, is it possible the linter isn't fully recursing?  I wonder if it's just missing some files.

tinybike: Re: file/table names, whatever you guys decide is fine with me, so long as the table names are consistent (i.e., either all camel case or all snake case, not some of each :grin: ).

src/blockchain/log-processors/designated-report-submitted.ts
epheph: 👍 i'm all for dumber code 😎

src/blockchain/log-processors/designated-report-submitted.ts
epheph: VSCode adds those spaces when you do the auto-format, i wonder if i can change that.

src/migrations/20171017150457_market_state.ts
epheph: I'll see about changing the whole thing to a knex-based create; seems like that might work and be the right solution anyway.

src/blockchain/log-processors/designated-report-submitted.ts
tinybike: I love me some dumb code! :grinning:

tinybike:
Aside from the (mostly minor / stylistic) issues raised above, this PR looks good.  Feel free to go ahead and merge it once the individual items are addressed!

src/migrations/20171015155157_reports_designated.ts
epheph: So i've migrated these tables to the createTable() format, but i don't see Knex supporting a "numeric" data type. We could go straight to decimal, which is more explicitly what we want, but then is that compatible/reasonable with pg?

src/blockchain/log-processors/designated-report-submitted.ts
tinybike: Hm, if you can't easily it's not a big deal, just style.

epheph:
Weird, opening up #19 resolved #17, since they are the exact same commit, but one is sourced from a remote fork, the other local.

PR #21, opened 2017-10-19
change to sorted array for get-markets-info

epheph:
https://app.clubhouse.io/augur/story/4254/getmarketsinfo-should-return-a-sorted-array-instead-of-an-object

### DIFF
--- a/src/server/getters/get-markets-info.ts
+++ b/src/server/getters/get-markets-info.ts
@@ -5,22 +5,19 @@ import { Address, MarketsRow, OutcomesRow, UIMarketInfo, UIMarketsInfo, UIOutcom
 import { reshapeOutcomesRowToUIOutcomeInfo, reshapeMarketsRowToUIMarketInfo } from "./get-market-info";
 
 export function getMarketsInfo(db: Knex, universe: Address|null|undefined, marketIDs: Array<Address>|null|undefined, callback: (err?: Error|null, result?: UIMarketsInfo) => void): void {
-  let query: string = "SELECT * FROM markets WHERE";
-  let queryParams: Array<Address|Array<Address>>;
+  let marketsQuery: Knex.QueryBuilder = db.select().from("markets").orderBy("creationTime");
+
   if (universe == null && marketIDs == null) {
     return callback(new Error("must include universe or marketIDs parameters"));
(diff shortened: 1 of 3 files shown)


src/server/getters/get-markets-info.ts
```diff
@@ -23,15 +21,16 @@ export function getMarketsInfo(db: Knex, universe: Address|null|undefined, marke
     query = `${query} "marketID" IN (??)`;
     queryParams = [marketIDs!];
   }
+  query = `${query} ORDER BY "creationTime" ASC`;
```
epheph: Not sure what the goal is in sorting this here. @stephensprinkle  do you know what order is desired for querying multiple markets at a time (like those that belong to a universe?)

src/server/getters/get-markets-info.ts
```diff
@@ -1,37 +1,31 @@
...
+    marketsQuery = marketsQuery.whereIn( "marketID", marketIDs );
+  }
+
+  marketsQuery.asCallback((err?: Error|null, marketsRows?: Array<MarketsRow>): void => {
     if (err) return callback(err);
     if (!marketsRows || !marketsRows.length) return callback(null);
-    const marketsInfo: UIMarketsInfo = {};
+    const marketsInfo: UIMarketsInfo = [];
     each(marketsRows, (marketsRow: MarketsRow, nextMarketsRow: ErrorCallback): void => {
       db.raw(`SELECT * FROM outcomes WHERE "marketID" = ?`, [marketsRow.marketID]).asCallback((err?: Error|null, outcomesRows?: Array<OutcomesRow>): void => {
```
pgebheim: Need to change this to the QB form as well or it won't return the right results against PGSql

pgebheim: Actually, take a look at writing this query as a join btw `markets` and `outcomes`.

epheph: Ha, can't believe i missed that raw call there, good catch.

epheph: it's a one-to-many table, it seems cleaner to use separate queries, or else we're returning the market over and over, collapsing it in the code, or using an aggregate sql function. Unless you know a better way with Knex?

PR #22, opened 2017-10-19
Bug/4274 oversyncing

pgebheim:
Change to using knex QB syntax so results are returned in a consistent format.

Do the subquery to find out if we're actually at the max synced.

### DIFF
--- a/src/blockchain/sync-augur-node-with-blockchain.ts
+++ b/src/blockchain/sync-augur-node-with-blockchain.ts
@@ -18,7 +18,6 @@ export function syncAugurNodeWithBlockchain(db: Knex,  augur: Augur, ethereumNod
         }).as("maxBlocks");
       }).asCallback( (err: Error|null, rows?: Array<HighestBlockNumberRow>): void => {
           if (err) return callback(err);
-          console.log(rows);
           const row: HighestBlockNumberRow = rows![0];
           const fromBlock: number = (!row || !row.highestBlockNumber) ? uploadBlockNumbers[augur.rpc.getNetworkID()] : row.highestBlockNumber + 1;
           const highestBlockNumber: number = parseInt(blockNumber, 16) - 1;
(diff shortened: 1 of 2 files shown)


src/blockchain/sync-augur-node-with-blockchain.ts
```diff
@@ -12,12 +12,16 @@ export function syncAugurNodeWithBlockchain(db: Knex,  augur: Augur, ethereumNod
...
     startAugurListeners(db, augur, (blockNumber: string): void => {
-      db.raw(`SELECT "highestBlockNumber" FROM blockchain_sync_history ORDER BY "highestBlockNumber" DESC LIMIT 1`)
-        .asCallback( (err: Error|null, rows?: Array<HighestBlockNumberRow>): void => {
+      db.max("block as highestBlockNumber").from(function(this: Knex.QueryBuilder): void {
+        this.max("highestBlockNumber as block").from("blockchain_sync_history").unionAll(function(this: Knex.QueryBuilder): void {
+          this.max("blockNumber as block").from("blocks");
+        }).as("maxBlocks");
+      }).asCallback( (err: Error|null, rows?: Array<HighestBlockNumberRow>): void => {
           if (err) return callback(err);
+          console.log(rows);
```
epheph: remove debug?

src/blockchain/sync-augur-node-with-blockchain.ts
```diff
@@ -12,12 +12,16 @@ export function syncAugurNodeWithBlockchain(db: Knex,  augur: Augur, ethereumNod
...
           const row: HighestBlockNumberRow = rows![0];
           const fromBlock: number = (!row || !row.highestBlockNumber) ? uploadBlockNumbers[augur.rpc.getNetworkID()] : row.highestBlockNumber + 1;
-          const highestBlockNumber: number = parseInt(blockNumber, 16);
+          const highestBlockNumber: number = parseInt(blockNumber, 16) - 1;
```
epheph: why `-1` ?

pgebheim: We receive block number N off the chain, and then sync the blocks before that which were unsynched.

```
Getting Augur logs from block 1377804 to block 1903664
new block: 1903665 1508440891
new block: 1903666 1508440893
```

PR #23, opened 2017-10-20
First pass at get-markets-awaiting-designated-reporting

epheph:
Writing tests now

### DIFF
--- a/src/server/getters/get-markets-awaiting-designated-reporting.ts
+++ b/src/server/getters/get-markets-awaiting-designated-reporting.ts
@@ -1,7 +1,24 @@
+import { each } from "async";
 import * as Knex from "knex";
-import { Address } from "../../types";
+import { Address, MarketsRow, UIMarketInfo, UIMarketsInfo, ErrorCallback } from "../../types";
+import { reshapeMarketsRowToUIMarketInfo } from "./get-market-info";
 
-// Look up all markets that are currently awaiting designated (automated) reporting. Should accept an designatedReporterAddress parameter that allows me to filter by designated reporter address.
+// Look up all markets that are currently awaiting designated (automated) reporting.
+// Should accept a designatedReporterAddress parameter that filters by designated reporter address.
 export function getMarketsAwaitingDesignatedReporting(db: Knex, designatedReporter: Address, callback: (err?: Error|null, result?: any) => void): void {
-
+    const marketsInfo: UIMarketsInfo = [];
+    let queryData: {} = {};
+    if ( designatedReporter != null ) {
+        queryData = { designatedReporter };
+    }
+    // TODO: should we also consider a market_state's phase IS NULL?
+    db.select().from("markets").where(queryData).whereNull("marketStateID").asCallback((err?: Error|null, marketsRows?: Array<MarketsRow>): void => {
+        if (err) return callback(err);
+        if (!marketsRows || !marketsRows.length) return callback(null);
+        each(marketsRows, (marketsRow: MarketsRow, nextMarketsRow: ErrorCallback): void => {
+            if (err) return callback(err);
+            marketsInfo.push(reshapeMarketsRowToUIMarketInfo(marketsRow, []) as UIMarketInfo);
+        });
+        callback(null, marketsInfo);
+    });
 }
--- a/src/server/getters/get-markets-awaiting-designated-reporting.ts
+++ b/src/server/getters/get-markets-awaiting-designated-reporting.ts
@@ -6,19 +6,12 @@ import { reshapeMarketsRowToUIMarketInfo } from "./get-market-info";
 // Look up all markets that are currently awaiting designated (automated) reporting.
 // Should accept a designatedReporterAddress parameter that filters by designated reporter address.
 export function getMarketsAwaitingDesignatedReporting(db: Knex, designatedReporter: Address|null, callback: (err?: Error|null, result?: any) => void): void {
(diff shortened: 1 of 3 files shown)


src/server/getters/get-markets-awaiting-designated-reporting.ts
```diff
@@ -1,7 +1,24 @@
+import { each } from "async";
 import * as Knex from "knex";
-import { Address } from "../../types";
+import { Address, MarketsRow, UIMarketInfo, UIMarketsInfo, ErrorCallback } from "../../types";
+import { reshapeMarketsRowToUIMarketInfo } from "./get-market-info";
 
-// Look up all markets that are currently awaiting designated (automated) reporting. Should accept an designatedReporterAddress parameter that allows me to filter by designated reporter address.
+// Look up all markets that are currently awaiting designated (automated) reporting.
+// Should accept a designatedReporterAddress parameter that filters by designated reporter address.
 export function getMarketsAwaitingDesignatedReporting(db: Knex, designatedReporter: Address, callback: (err?: Error|null, result?: any) => void): void {
```
pgebheim: if `designatedReporter` can be null, specify this in the type `designatedReporter: Addess | null`

src/server/getters/get-markets-awaiting-designated-reporting.ts
```diff
@@ -1,7 +1,24 @@
...
+    const marketsInfo: UIMarketsInfo = [];
+    let queryData: {} = {};
+    if ( designatedReporter != null ) {
+        queryData = { designatedReporter };
+    }
+    // TODO: should we also consider a market_state's phase IS NULL?
+    db.select().from("markets").where(queryData).whereNull("marketStateID").asCallback((err?: Error|null, marketsRows?: Array<MarketsRow>): void => {
+        if (err) return callback(err);
+        if (!marketsRows || !marketsRows.length) return callback(null);
+        each(marketsRows, (marketsRow: MarketsRow, nextMarketsRow: ErrorCallback): void => {
```
pgebheim: Do you really wanna use `async.each` here?

If so,  make sure to call `nextMarketsRow`

src/server/getters/get-markets-awaiting-designated-reporting.ts
```diff
@@ -1,7 +1,24 @@
...
+            if (err) return callback(err);
+            marketsInfo.push(reshapeMarketsRowToUIMarketInfo(marketsRow, []) as UIMarketInfo);
```
pgebheim: No type cast needed, `reshape...` returns UIMarketInfo

src/server/getters/get-markets-awaiting-designated-reporting.ts
MicahZoltu: Unrelated to the above comment, but it caused me to look at this code:

If you control the callsite or can influence its behavior, I recommend either `apple?: type` or `apple: type|null`.  `apple?: type|null` is ultra-permissive which increases the chance of bugs.  In general, if possible I recommend constraining accepted types as much as possible.  (commenting on `err?: Error|null` which is one step away from `any`)

src/server/getters/get-markets-awaiting-designated-reporting.ts
tinybike: I think I added a few of these in places where they're unnecessary (b/c I'm a noob), please feel free to delete em!

src/server/getters/get-markets-awaiting-designated-reporting.ts
tinybike: I *think* we could generally use `err: Error|null`.  We're using the standard Node-style callbacks, so callbacks are always either `callback(err)` or `callback(null, result)`.  Can built-in knex functions call callbacks with the error argument _missing_?  (@pgebheim would know better than me)

Either way, since we'd be updating the error callback signatures in a ton of different functions, that'd be a separate PR if we want to go that route.

epheph:
Thanks for the feedback @pgebheim , addressed

epheph:
How do we want to handle returning data that is normally a part of a market, but doesn't make sense if it is not reported? 
```
"consensus": null,
"finalizationTime": null
```
is how it is currently handled

src/server/getters/get-markets-awaiting-designated-reporting.ts
```diff
@@ -15,9 +14,10 @@ export function getMarketsAwaitingDesignatedReporting(db: Knex, designatedReport
     db.select().from("markets").where(queryData).whereNull("marketStateID").asCallback((err?: Error|null, marketsRows?: Array<MarketsRow>): void => {
         if (err) return callback(err);
         if (!marketsRows || !marketsRows.length) return callback(null);
-        each(marketsRows, (marketsRow: MarketsRow, nextMarketsRow: ErrorCallback): void => {
+        const marketsInfo: UIMarketsInfo = [];
+        marketsRows.map((marketsRow: MarketsRow): void => {
```
pgebheim: `map` should return the value it implicitly creates an array.

```typescript
% [1, 2, 3].map((v) => v+10)
=> [11, 12, 13]
```

So in this case

```typescript
const marketsInfo: UIMarketsInfo = marketsRows.map(_.partialRight(reshapeMarketsRowToUIMarketInfo, []));
```
Will get you what you want, will also need to import lodash at the top for `_.partialRight`

```typescript
import * as _ from "lodash";
```

PR #26, opened 2017-10-20
Lodashify @jack's code

pgebheim:
More declarativeness.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,12 @@
     "augur.js": "4.4.10",
     "bignumber.js": "4.1.0",
     "knex": "0.13.0",
+    "lodash.get": "^4.4.2",
+    "lodash.groupby": "^4.6.0",
     "lodash.isnull": "3.0.0",
     "lodash.merge": "^4.6.0",
     "lodash.omit": "4.5.0",
+    "lodash.partial": "^4.2.1",
     "pg": "7.3.0",
     "postinstall-build": "^5.0.1",
     "sqlite3": "3.1.8",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,9 +44,12 @@
     "augur.js": "4.4.10",
     "bignumber.js": "4.1.0",
     "knex": "0.13.0",
+    "lodash.get": "^4.4.2",
```
tinybike: Should use fixed version numbers.  (We're amassing quite a collection here...)

src/server/getters/get-market-price-history.ts
```diff
@@ -1,31 +1,30 @@
 import * as Knex from "knex";
+import * as _ from "lodash";
 import { Address, TimestampedPrice, MarketPriceHistory } from "../../types";
 import { sortDirection } from "../../utils/sort-direction";
 
-interface MarketPriceHistoryTradesRow {
+interface MarketPriceHistoryTradesRow extends TimestampedPrice {
```
tinybike: Hm, if anything I think the conceptual relationship is the other way -- TimestampedPrice is derived from TradesRow.  That said, I'd sure prefer to avoid inheritance if we can, even if that means we've got some duplicate field names etc.  Does TS have a way to do interface composition?

src/server/getters/get-market-price-history.ts
```diff
@@ -1,31 +1,30 @@
...
-    tradesRows.forEach((trade: MarketPriceHistoryTradesRow): void => {
-      if (!marketPriceHistory[trade.outcome]) marketPriceHistory[trade.outcome] = [];
-      marketPriceHistory[trade.outcome].push({
-        price: trade.price,
-        timestamp: trade.tradeTime
-      });
-    });
+
+    // Group by outcome, drop "outcome" from result
+    const marketPriceHistory: MarketPriceHistory = _.mapValues(
```
tinybike: This is more declarative, but IMO less obvious / a bit harder to read.  Do you have a strong preference for this style?  If not, my vote is to stick with the old, dumb code -- remember that I have a great fondness for dumb code, the dumber the better! ;)

tinybike:
Gonna close this.  Sorry for "mixed signals" on this one @pgebheim -- I didn't realize how much I preferred the original dumb-as-mud version until I saw the update!

PR #27, opened 2017-10-21
4310 linty mclintface

tinybike:
I'm not clear on *why* this is necessary, but apparently it is...


tinybike:
I think this is pretty straightforward / improves linting so I'm gonna go ahead and merge.

PR #28, opened 2017-10-21
First pass at supporting managing subscriptions in Augur Node

pgebheim:
Give feedback!


package.json
```diff
@@ -40,6 +40,7 @@
   "dependencies": {
     "@types/knex": "0.0.61",
     "@types/lodash": "4.14.77",
+    "@types/uuid": "^3.4.3",
```
tinybike: Fix version numbers.

package.json
```diff
@@ -50,6 +51,7 @@
     "pg": "7.3.0",
     "postinstall-build": "^5.0.1",
     "sqlite3": "3.1.8",
+    "uuid": "^3.1.0",
```
tinybike: Fix version numbers.

package.json
```diff
@@ -50,6 +51,7 @@
     "pg": "7.3.0",
     "postinstall-build": "^5.0.1",
```
tinybike: Fix version numbers.

src/server/run-websocket-server.ts
```diff
@@ -1,29 +1,54 @@
...
+          const eventName: string = message.params.shift();
+
+          try {
+            const subscription: string = subscriptions.subscribe(eventName, message.params, (data: {}): void => {
+              websocket.send(makeJsonRpcResponse(message.id, { subscription, result: data}));
+            });
+            websocket.send(makeJsonRpcResponse(message.id, { subscription }));
+          } catch (exc) {
+            console.error("suscription error", exc, data);
+            websocket.send(makeJsonRpcResponse(message.id, false));
```
tinybike: We should probably have a separate `makeJsonRpcError` function that creates the error response in the format of http://www.jsonrpc.org/specification#error_object

src/server/run-websocket-server.ts
```diff
@@ -1,29 +1,54 @@
...
       try {
         const message = JSON.parse(data as string);
-        if (isJsonRpcRequest(message)) {
+        if (!isJsonRpcRequest(message)) return console.error("bad json rpc message received:", message);
+
+        if (message.method === "subscribe") {
+          const eventName: string = message.params.shift();
+
+          try {
+            const subscription: string = subscriptions.subscribe(eventName, message.params, (data: {}): void => {
```
tinybike: I'm a little unclear what `data` is expected to be here -- is this just the data attached to `emit`, so it can have a pretty arbitrary shape?  Also, what is `message.id` in this context?  Since we don't have a request `id` to match the response with, should it just be `null` for messages pushed to subscribers?

src/server/run-websocket-server.ts
```diff
@@ -1,29 +1,54 @@
...
+              websocket.send(makeJsonRpcResponse(message.id, { subscription, result: data}));
```
tinybike: Mismatched brace spacing.

src/server/run-websocket-server.ts
```diff
@@ -1,29 +1,54 @@
...
+    const subscriptions = new Subscriptions();
+
     websocket.on("message", (data: WebSocket.Data): void => {
       try {
         const message = JSON.parse(data as string);
-        if (isJsonRpcRequest(message)) {
+        if (!isJsonRpcRequest(message)) return console.error("bad json rpc message received:", message);
+
+        if (message.method === "subscribe") {
+          const eventName: string = message.params.shift();
```
tinybike: Are there other parameters expected / allowed here, besides the event name?

One important thing to note is that there can be duplicate event names across different contracts; e.g., all ERC-20 contracts have their own `Transfer` and `Approval` events.  Would it be better to key the subscriptions by contract name and event name?  (This is the way augur.js handles event subscriptions internally.)

src/server/subscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.emit("removeAllListeners");
+    return super.removeAllListeners(eventName);
+  }
+
+  private subscribeToEvent(eventName: string, params: any, publish: (data: {}) => void): string {
+    const subscription: string = uuidv4();
+
+    const handler = (data: {}): void => {
+      this.emit(eventName, data);
+    };
```
tinybike: I think you need to either `.bind(this)` or (maybe cleaner?) inline the `this.emit` call.  i.e.,
```
const handler = (data: {}): void => { this.emit(eventName, data); }.bind(this);
```
or
```
const handler = (data: {}): void => this.emit(eventName, data);
```

src/server/subscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    augurEmitter.on(eventName, handler);
+
+    this.on(eventName, publish);
+
+    // Unsubscribe from one subscription
+    this.once(`unsubscribe:${publish}`, (): void => {
```
tinybike: I think this should be `unsubscribe:${subscription}` instead of `publish`.

src/server/subscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+      this.removeListener(eventName, publish);
+      augurEmitter.removeListener(eventName, handler);
+    });
```
tinybike: Same comment as w/ `handler`, I think you need to `.bind(this)` to the callback or `this.removeListener` will not work (`this` will be undefined).

src/server/subscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+      default:
+        throw new Error(`Event ${eventName} not available for subscription`);
+    }
+  }
+
+  public unsubscribe(subscription: string): void  {
+    this.emit(`unsubscribe:${subscription}`);
+  }
+
+  public removeAllListeners(eventName?: string | symbol | undefined): this {
```
tinybike: Is `this` the correct/intended return type for `removeAllListeners`?

src/server/run-websocket-server.ts
```diff
@@ -1,29 +1,54 @@
...
-        } else {
-          console.error("bad json rpc message received:", message);
         }
        } catch (exc) {
          console.error("bad json rpc message received:", exc, data);
        }
     });
+
+    websocket.on("close", () => {
+      subscriptions.removeAllListeners();
```
tinybike: Design question: would it be possible/desirable to allow for reconnection attempts before removing all listeners?

src/server/subscriptions.ts
tinybike: Also, when will the `eventName` parameter be included vs left out for this function?  If this simply removes all listeners, is the event name param needed at all?

src/server/run-websocket-server.ts
pgebheim: Good call!

src/server/run-websocket-server.ts
pgebheim: 1) "data" is just whatever payload we push down the pipe to a subscriber. So in our case it would likely be the database row(s) that we're committing when that event occurs, or some subset of that data.

e.g. for `MarketCreated` I just do `augurEmitter.emit('MarketCreated', dataToInsert);` which corresponds to the row data.

2) Messages pushed to subscribers have no message.id. The immediate response to the subscribe/unsubscribe calls. `message.id` in all the contexts is a Request<->Response mapping which is only valid per connection.

src/server/run-websocket-server.ts
pgebheim: Do we have a use case for needing to specify the contract? I was under the assumption that we were working on these just from the "Augur" perspective, so we're subscribing to MarketCreated as an augur construct, not an arbitrary log.

src/server/subscriptions.ts
pgebheim: Fat arrow syntax in TS/ES6 will automatically bind `this` to the the outer scope. But yes to putting it on one line!

src/server/subscriptions.ts
pgebheim: Lol, yes yes yes :-D

src/server/subscriptions.ts
pgebheim: Again, ES Fat-Arrow will preserve "this" without using bind.

Internally it will transpile to something like

```
var self = this;
this.once(`unsubscribe:${eventName}`, function() {
  self.removeListener(eventName, publish);
  ...
});
```

src/server/subscriptions.ts
pgebheim: This is the protoype for EventEmitter#removeAllListeners, and yes the return type is `this` -- it works out to be the current class type because in JS a class is a function, and the function definitions are within that context. Weird, yes, but blame it on the prototype...

removeAllListeners can remove all listeners on the Emitter or all listeners for a specific key. This is because you can bind multiple listeners to one event for an event emitter.

src/server/run-websocket-server.ts
pgebheim: Possible, yes, desirable -- probably not.

To manage this sort of reconnection logic it requires more bookkeeping on the client and the server.

I think it is in general cleaner to require that we re-bind all listeners after the connection has happened on the client side.

Doing it this way does mean we are way more likely to drop events on the floor in the disconnect/reconnect cycle, and requires that we handle that case if it's an issue.

src/server/run-websocket-server.ts
pgebheim: And yes, using the same format as geth for subscriptions https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB.

Ultimately the API supports passing parameters into the subscriptions e.g. to further filter the events delivered.

My internal API supports this, although I don't think we use it for anything.

src/server/run-websocket-server.ts
pgebheim: I'll take this chance to translate all these errors to the spec.

src/server/subscriptions.ts
pgebheim: leaving the braces so we throw away the return value

src/server/subscriptions.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.emit("removeAllListeners");
+    return super.removeAllListeners(eventName);
+  }
+
+  private subscribeToEvent(eventName: string, params: any, publish: (data: {}) => void): string {
+    const subscription: string = uuidv4();
+
+    const handler = (data: {}): void => { this.emit(eventName, data) };
```
epheph: lint, missing `;`

src/server/run-websocket-server.ts
```diff
@@ -1,30 +1,60 @@
...
+            websocket.send(makeJsonRpcResponse(message.id, { subscription }));
+          } catch (exc) {
+            websocket.send(makeJsonRpcError(message.id, JsonRpcErrorCode.MethodNotFound, exc.toString(), false));
+          }
+        } else if (message.method === "unsubscribe") {
+          const subscription: string = message.params.shift();
+          subscriptions.unsubscribe(subscription);
+          websocket.send(makeJsonRpcResponse(message.id, true));
+        } else {
+          dispatchJsonRpcRequest(db, message as JsonRpcRequest, (err?: Error | null, result?: any): void => {
```
epheph: In another PR, we changed most of the `err?` to `err`, couldn't find any instances where it would be `undefined`

src/server/subscriptions.ts
tinybike: Right, I use `this` so rarely that I wasn't even aware of this fat-arrow function feature...

test/server/subscriptions.js
```diff
@@ -0,0 +1,86 @@
+"use strinct";
```
tinybike: Typo

tinybike:
Looks good, merging.

PR #31, opened 2017-10-23
3852 user trading positions

tinybike:
Added a positions table, filled in the OrderFilled log processor, filled in the getUserTradingPositions getter.


src/blockchain/log-processors/order-filled.ts
```diff
@@ -1,17 +1,167 @@
...
+    forEachOf(positionInMarket, (numShares: string, outcome: number, nextOutcome: AsyncCallback): void => {
+      calculateProfitLossInOutcome(augur, trx, account, marketID, outcome, (err: Error|null, profitLossInOutcome?: any): void => {
+        if (err) return nextOutcome(err);
+        const { realized, unrealized, position } = profitLossInOutcome!;
+        realizedProfitLoss[outcome] = realized;
+        unrealizedProfitLoss[outcome] = unrealized;
+        positionInMarketAdjustedForUserIntention[outcome] = position;
+        nextOutcome();
+      });
+    }, (err?: any): void => {
```
epheph: `err: Error|null` ?

src/blockchain/log-processors/market-created.ts
```diff
@@ -30,10 +32,10 @@ export function processMarketCreatedLog(db: Knex, augur: Augur, trx: Knex.Transa
           marketCreator:              log.marketCreator,
           creationBlockNumber:        log.blockNumber,
           creationFee:                log.marketCreationFee,
-          creationTime:               blocksRows![0]!.blockTimestamp,
+          creationTime:               timestamp,
           marketType:                 extraInfo!.marketType,
-          minPrice:                   extraInfo!.minPrice,
-          maxPrice:                   extraInfo!.maxPrice,
+          minPrice:                   new BigNumber(extraInfo!.minPrice, 10).toFixed(),
```
epheph: This converts a base-10 string to a base-10 string, is it to cut leading zeroes or to make sure that there is no decimal?

src/blockchain/log-processors/order-filled.ts
```diff
@@ -1,17 +1,167 @@
...
+    }).asCallback(nextOutcome);
+  }, callback);
+}
+
+export function calculateProfitLossInOutcome(augur: Augur, trx: Knex.Transaction, account: Address, marketID: Address, outcome: number, callback: (err: Error|null, profitLossInOutcome?: any) => void): void {
+  getUserTradingHistory(trx, account, marketID, outcome, null, null, null, null, null, (err: Error|null, userTradingHistory?: Array<UITrade>): void => {
+    if (err) return callback(err);
+    trx.select("price").from("outcomes").where({ marketID, outcome }).asCallback((err: Error|null, outcomesRows?: Array<OutcomesRow>): void => {
+      if (err) return callback(err);
+      const { price } = outcomesRows![0];
```
epheph: Should there be a check here for `!outcomesRows || !outcomesRows.length` ?

src/blockchain/log-processors/order-filled.ts
```diff
@@ -1,17 +1,167 @@
...
+            amount: (next: AsyncCallback): void => augur.api.Orders.getAmount(orderID, next),
+            creatorPositionInMarket: (next: AsyncCallback): void => augur.trading.getPositionInMarket({ market: marketID, address: log.creator }, next),
+            fillerPositionInMarket: (next: AsyncCallback): void => augur.trading.getPositionInMarket({ market: marketID, address: log.filler }, next),
+          }, (err, onContractData): void => {
+            if (err) return callback(err);
+            const { amount, creatorPositionInMarket, fillerPositionInMarket } = onContractData!;
+            const amountRemainingInOrder = new BigNumber(amount!, 16).toFixed();
+            let updateOrdersQuery = db("orders").transacting(trx).where({ orderID });
+            if (amountRemainingInOrder === "0") {
+              updateOrdersQuery = updateOrdersQuery.del();
```
epheph: We'll need to think about what to do here when we implement blockstream reorgs

src/blockchain/log-processors/order-filled.ts
```diff
@@ -1,17 +1,167 @@
...
+
+export function upsertPositionInMarket(db: Knex, augur: Augur, trx: Knex.Transaction, account: Address, marketID: Address, positionInMarket: Array<string>, callback: ErrorCallback): void {
+  trx.select("outcome").from("positions").where({ account, marketID }).asCallback((err: Error|null, positionsRows?: Array<PositionsRow>): void => {
+    if (err) return callback(err);
+    const numOutcomes = positionInMarket.length;
+    const realizedProfitLoss = new Array(numOutcomes);
+    const unrealizedProfitLoss = new Array(numOutcomes);
+    const positionInMarketAdjustedForUserIntention = new Array(numOutcomes);
+    forEachOf(positionInMarket, (numShares: string, outcome: number, nextOutcome: AsyncCallback): void => {
+      calculateProfitLossInOutcome(augur, trx, account, marketID, outcome, (err: Error|null, profitLossInOutcome?: any): void => {
```
epheph: `profitLossInOutcome` could benefit from a type

tinybike: True -- this should be an augur.js type actually, I can add that.

src/blockchain/log-processors/order-filled.ts
tinybike: Yeah :grimacing: -- I think the log processors are basically just gonna need equal-and-opposite "log de-processors" that reverse the changes -- I can't think of a way to do it that's more general / doesn't require special-purpose logic for each log processor.

src/blockchain/log-processors/order-filled.ts
tinybike: Hm, I'm not 100% clear what all the combinations of things that can be returned from the knex methods are -- but yes that'd be good defensive practice.  I'll add!

src/blockchain/log-processors/market-created.ts
tinybike: Oh, good point -- the extraInfo fields aren't base-16.  There shouldn't be a need for this conversion at all then, I will remove.

src/blockchain/log-processors/order-filled.ts
epheph: I think that actually deleting a row makes it more complicated (impossible maybe?). If we kept state updates in separate tables, and linked via primary key, we could just  re-link the the prior version (and delete the now-orphaned row in the linked table). That's what I did with `market_state` table. I could imagine some helper functions that acted on the two tables fairly generically, but yea, i can't picture a more automated way of accomplishing that.

src/blockchain/log-processors/order-filled.ts
epheph: Yea, i'd love to take some time and play around with Knex to see what values it callback's with in various circumstances! It might never be undefined/null here, and we could clean up a lot of code.

PR #32, opened 2017-10-23
Consolidate markets awaiting reporting

epheph:
The gist of this PR is: `getMarketsAwaitingLimitedReporting` and `getMarketsAwaitingAllReporting` are merged into a single getter, `getMarketsAwaitingReporting` with an argument for `reportingState`


epheph:
@tinybike @pgebheim  OK this is ready for initial review. I've renamed `phase` to `reportingState`, it doesn't make sense to make up new terms for what is effectively the same thing. 
Before merging, i would like to add:

1.) an enum for reportingState (plus maybe a lookup table? not sure if that's worthwhile) to match exactly this:
https://github.com/AugurProject/augur-core/blob/43d4b43779bd12cbb13d78708775ebb9e510812f/source/contracts/reporting/IMarket.sol#L15

2.) real test cases

epheph:
@stephensprinkle how do you want to interact with reportingState? It is a field of a market, as well as something you can filter on. Do you want to use a shared enum? A string?

src/server/getters/get-market-info.ts
```diff
@@ -52,8 +53,16 @@ export function reshapeMarketsRowToUIMarketInfo(row: MarketsRow, outcomesInfo: A
   return marketInfo;
 }
 
+export function getMarketsWithReportingState(db: Knex, selectColumns?: Array<string>): Knex.QueryBuilder {
```
epheph: @pgebheim thoughts about the right place to put this? Feels funny to put it in another getter's file, but other meta-query-related things were already here and included from other getters.

tinybike: I agree having these meta-getters and reorganization functions inside the getter file (and/or equivalent things inside log-processor files) feels "off"... Maybe we should put them in a separate "helpers" directory?  Another (maybe more consistent with our overall organization) option is, we could change get-market-info.ts into a directory with the main getMarketInfo function as index.ts, and one helper function per file.

For now I think it's fine to put it here in the get-market-info.ts file though!

src/server/getters/get-market-info.ts
```diff
@@ -52,8 +53,16 @@ export function reshapeMarketsRowToUIMarketInfo(row: MarketsRow, outcomesInfo: A
...
+  // TODO: turn leftJoin() into join() once we take care of market_state on market creation
+  const columns = selectColumns ? selectColumns.slice() : ["markets.*"];
+  return db.select(columns.concat("market_state.reportingState"))
+    .from("markets")
+    .leftJoin("market_state", "markets.marketStateID", "market_state.marketStateID");
```
tinybike: Hm, is left join the correct operation here?  This will use all rows from `markets`, even ones that don't have a matching entry in `market_state`.

src/server/getters/get-markets-awaiting-designated-reporting.ts
```diff
@@ -1,14 +1,14 @@
...
-import { reshapeMarketsRowToUIMarketInfo } from "./get-market-info";
+import { reshapeMarketsRowToUIMarketInfo, getMarketsWithReportingState } from "./get-market-info";
 import { sortDirection } from "../../utils/sort-direction";
 
 // Look up all markets that are currently awaiting designated (automated) reporting.
 // Should accept a designatedReporterAddress parameter that filters by designated reporter address.
 export function getMarketsAwaitingDesignatedReporting(db: Knex, designatedReporter: Address|null, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err?: Error|null, result?: any) => void): void {
   const queryData: {} = (designatedReporter != null) ? { designatedReporter } : {};
-  let query = db("markets").where(queryData).whereNull("marketStateID");
+  let query = getMarketsWithReportingState(db).where(queryData).whereNull("markets.marketStateID");
```
tinybike: Hm, possibly would be clearer to write this as --
```
let query = getMarketsWithReportingState(db).whereNull("markets.marketStateID");
if (designatedReporter != null) query = query.where({ designatedReporter });
```
What do you think?

src/server/getters/get-markets-awaiting-reporting.ts
```diff
@@ -0,0 +1,26 @@
+import * as Knex from "knex";
+import { Address, MarketsRow, UIMarketInfo, UIMarketsInfo, ErrorCallback } from "../../types";
+import { reshapeMarketsRowToUIMarketInfo, getMarketsWithReportingState } from "./get-market-info";
+import { sortDirection } from "../../utils/sort-direction";
+
+// Look up all markets that are currently available for limited reporting.
+// Must be able to sort by number of reports submitted for each market so far, and the response should include the number of reports already submitted -- as well as the payoutNumerator values of each of the reports + the amount staked on each -- as part of the response.
+export function getMarketsAwaitingReporting(db: Knex, reportingWindow: Address|null, reportingRound: number|null, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err: Error|null, result?: any) => void): void {
```
tinybike: All of the getMarkets* functions (except getMarketsInfo) should return a simple array of market IDs, rather than the big complicated `UIMarketsInfo` object.

(Unless @stephensprinkle has requested otherwise, in which case please do as he asks!)

test/server/getters/get-markets-awaiting-reporting.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+const unlink = require("fs").unlink;
+const join = require("path").join;
```
tinybike: Note: `unlink` and `join` are both unused (I know this is just cruft from other test files, my bad for not removing these everywhere...).

test/server/getters/get-markets-awaiting-reporting.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+  };
+  test({
+    description: "get markets awaiting unknown reportingWindow",
+    params: {
+      reportingWindow: "0xf0f0f0f0f0f0f0f0b0b0b0b0b0b0b0f0f0f0f0b0"
+    },
+    assertions: (err, marketsAwaitingReporting) => {
+      assert.isNull(err);
+      assert.deepEqual(marketsAwaitingReporting, [])
```
tinybike: Missing semicolon.

src/server/getters/get-market-info.ts
epheph: yes, only until we implement market_state creation for new markets. without this, if a row in market_state didn't exist, it wouldn't be returned. The TODO above is so we don't forgot to remove it once we fully implement market_state

src/server/getters/get-markets-awaiting-reporting.ts
epheph: sure thing, easy to change. I'll wait to hear from @stephensprinkle

tinybike: Note that if he hasn't already said otherwise regarding this particular getter, he has expressed that the UI prefers that *all* getters, as a general rule, simply return arrays of market IDs.  So I think that's the desired behavior here as well.

src/server/getters/get-markets-awaiting-designated-reporting.ts
epheph: That's much better. I think you could argue the queryData object works if you can do the build in a single line, but if you're making it a variable anyway, this is significantly better

test/server/getters/get-markets-awaiting-reporting.js
epheph: Huh, does the linter not check `test/` ? fixed

test/server/getters/get-markets-awaiting-reporting.js
epheph: removed from all tests

src/server/getters/get-markets-awaiting-reporting.ts
epheph: 👍

test/server/getters/get-markets-awaiting-reporting.js
tinybike: No, it doesn't at the moment, because we use tslint but the tests are written in regular JS.  Want to set up an .eslintrc just for linting the tests?  Would probably be useful!

src/server/getters/get-markets-awaiting-reporting.ts
epheph: how do we feel about comma-dangle in javascript? I copied from another augur project, but it didn't have comma-dangle as `always-multiline`, do we want to mimic the same behavior as tslint?

tinybike: Yes, I think `always-multiline` is a good idea!

tinybike:
@epheph The getMarkets* methods, apart from getMarketsInfo, should *not* return the big fancy UIMarketInfo objects; instead they should just return an array of market IDs.  Once this change is made, I think this PR should be ready-to-go.

PR #33, opened 2017-10-24
be consistent about empty array vs. undefined in response

epheph:
Small change, but it's how I implemented #32 , so I wanted to make it consistent.


tinybike:
Looks good.

PR #34, opened 2017-10-25
Adds log removal support

tinybike:
Added log removal support for Approval, Transfer, MarketCreated, OrderCanceled, OrderCreated, and OrderFilled events.  @epheph let me know you think if this looks like a reasonable approach to log removal!

### DIFF
--- a/src/blockchain/log-processors/legacy-rep-contract/transfer.ts
+++ b/src/blockchain/log-processors/legacy-rep-contract/transfer.ts
@@ -3,15 +3,17 @@ import * as Knex from "knex";
 import { FormattedLog, ErrorCallback } from "../../../types";
 
 export function processTransferLog(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
-  const dataToInsert: {} = {
-    transactionHash:  log.transactionHash,
-    logIndex:         log.logIndex,
-    sender:            log.from,
-    recipient:         log.to,
-    token:             log.address,
-    value:             log.value,
-    blockNumber:      log.blockNumber,
-  };
+  db.transacting(trx).insert({
+    transactionHash: log.transactionHash,
+    logIndex:        log.logIndex,
+    sender:          log.from,
+    recipient:       log.to,
+    token:           log.address,
+    value:           log.value,
+    blockNumber:     log.blockNumber,
+  }).into("transfers").asCallback(callback);
+}
 
-  db.transacting(trx).insert(dataToInsert).into("transfers").asCallback(callback);
+export function processTransferLogRemoval(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
+  db.transacting(trx).from("transfers").where({ transactionHash: log.transactionHash, logIndex: log.logIndex, blockNumber: log.blockNumber }).del().asCallback(callback);
 }
--- a/src/blockchain/log-processors/legacy-rep-contract/transfer.ts
+++ b/src/blockchain/log-processors/legacy-rep-contract/transfer.ts
@@ -15,5 +15,5 @@ export function processTransferLog(db: Knex, augur: Augur, trx: Knex.Transaction
 }
 
 export function processTransferLogRemoval(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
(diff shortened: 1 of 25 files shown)


src/blockchain/log-processors/order-created.ts
```diff
@@ -56,8 +56,20 @@ export function processOrderCreatedLog(db: Knex, augur: Augur, trx: Knex.Transac
...
+            db.transacting(trx).from("orders").where(orderID).update(orderData).asCallback(callback);
+          }
+        });
       });
     });
   });
 }
+
+export function processOrderCreatedLogRemoval(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
+  processOrderCanceledLog(db, augur, trx, log, callback);
```
epheph: Interesting, so we're not actually removing a created order during a reorg, we're just marking it as cancelled, but when it comes along to be recreated later, the insert-or-update logic above will do the right thing? The artifact here is that an order will exist with `isRemoved: 1`, but that seems harmless

epheph:
Using blockstream, we can also get notified of block removals as well, what do you think of implementing that? The current `onNewBlock()` code is inelegant (using `db.raw`). I took a pass at doing a quick fix, but will need to implement real revert detection, not just `ON CONFLICT ON CONSTRAINT` or the like. Here's the code from the cleanup, (but does not handle reorgs):

https://github.com/AugurProject/augur-node/compare/sbigelow/block-table-knex?expand=1

src/blockchain/log-processors/order-created.ts
tinybike: Yep, that seemed to be the easiest way to deal with the whole "but if we cancel it and the cancel gets removed, we no longer have the order info so what do we do :(" problem!

src/blockchain/log-processors/legacy-rep-contract/transfer.ts
```diff
@@ -3,15 +3,17 @@ import * as Knex from "knex";
...
+    recipient:       log.to,
+    token:           log.address,
+    value:           log.value,
+    blockNumber:     log.blockNumber,
+  }).into("transfers").asCallback(callback);
+}
 
-  db.transacting(trx).insert(dataToInsert).into("transfers").asCallback(callback);
+export function processTransferLogRemoval(db: Knex, augur: Augur, trx: Knex.Transaction, log: FormattedLog, callback: ErrorCallback): void {
+  db.transacting(trx).from("transfers").where({ transactionHash: log.transactionHash, logIndex: log.logIndex, blockNumber: log.blockNumber, token: log.address }).del().asCallback(callback);
```
epheph: Could just use transactionHash/logIndex ?

tinybike:
Ah, I'm not sure if BlockStream emits a special notification for blocks when reorgs happen (@MicahZoltu would prob know).  I think what we'd need to do is, if we ever see a duplicate block number (meaning that a reorg has happened), drop all blocks that are _newer_ than the duplicate block.  The other tables are all updated off of the log notifications, so they shouldn't need manual updating in this case.

src/blockchain/log-processors/legacy-rep-contract/transfer.ts
tinybike: That seems reasonable.  Is that guaranteed to be unique?

epheph:
```CREATE TABLE transfers (
      "transactionHash" varchar(66) NOT NULL,
      "logIndex" integer NOT NULL CONSTRAINT "nonnegativeLogIndex" CHECK ("logIndex" >= 0),
      sender varchar(66) NOT NULL,
      recipient varchar(66) NOT NULL,
      token varchar(66) NOT NULL,
      value numeric NOT NULL CONSTRAINT "positiveValue" CHECK (value > 0),
      "blockNumber" integer NOT NULL CONSTRAINT "positiveBlockNumber" CHECK ("blockNumber" > 0),

      UNIQUE("transactionHash", "logIndex")

    );
```

epheph:
@MicahZoltu and i were discussing this yesterday, it does emit block removal notifications.

tinybike:
Oh nice!  Then it should be simple to just write an `onBlockRemoved` function that removes blocks.  Is it the same as the log notifications, where there's a `removed: true` field attached to the block, or how's that work?

tinybike:
Yes, our types in general could use a bit of cleanup, I think there's a fair amount of duplication etc. as well.

PR #38, opened 2017-10-27
Handle block removals

tinybike:
Added processBlockRemoval method.  Also cleaned up onNewBlock and renamed to processBlock.


src/blockchain/process-block.ts
```diff
@@ -0,0 +1,47 @@
...
+    const timestamp = parseInt(block.timestamp, 16);
+    console.log("new block:", blockNumber, timestamp);
+    db.transaction((trx: Knex.Transaction): void => {
+      trx("blocks").where({ blockNumber }).asCallback((err: Error|null, blocksRows?: Array<BlocksRow>): void => {
+        if (err) return logError(err);
+        let query: Knex.QueryBuilder;
+        if (!blocksRows || !blocksRows.length) {
+          query = db.transacting(trx).insert({ blockNumber, timestamp }).into("blocks");
+        } else {
+          query = db("blocks").transacting(trx).where({ blockNumber }).update({ timestamp });
```
epheph: Is this logic required any more? Seems like if we are appropriately removing blocks, there's no need for an UPSERT?

src/blockchain/process-block.ts
```diff
@@ -0,0 +1,47 @@
...
+          } else {
+            trx.commit();
+          }
+        });
+      });
+    });
+  });
+}
+
+export function processBlockRemoval(db: Knex, blockNumberString: string): void {
```
epheph: It's the same thing, but should we change the type to match here, Int256? to match the declaration in start-augur-listeners?

src/blockchain/process-block.ts
tinybike: That's a good point -- I think this is needed until Paul's fixes to initial sync (using blockstream etc.) go in because the way things are now, we can technically get duplicate block notifications (from initial sync, then again from block listener).

src/blockchain/process-block.ts
tinybike: Ah, good point, yes we should!

PR #40, opened 2017-10-27
centralize join logic for markets

epheph:
This is a small change now, but I imagine one of the sorting options we're going to want to have is `creationTime`? (even though technically, `creationBlockNumber` will give same sort order)? And that would mean that join would have to happen in all getters, not just those that return full details. I imagine we'll join more tables and want to start following a similar pattern to prevent the join logic from spreading to every getter.

### DIFF
--- a/src/server/getters/get-markets-info.ts
+++ b/src/server/getters/get-markets-info.ts
@@ -2,7 +2,7 @@ import { each } from "async";
 import BigNumber from "bignumber.js";
 import * as Knex from "knex";
 import { Address, MarketsRowWithCreationTime, OutcomesRow, UIMarketInfo, UIMarketsInfo, UIOutcomeInfo, ErrorCallback } from "../../types";
-import { reshapeOutcomesRowToUIOutcomeInfo, reshapeMarketsRowToUIMarketInfo, getMarketsWithReportingState } from "./get-market-info";
+import { reshapeOutcomesRowToUIOutcomeInfo, reshapeMarketsRowToUIMarketInfo, getMarketsWithReportingState } from "./database";
 import { queryModifier } from "./database";
 
 export function getMarketsInfo(db: Knex, universe: Address|null|undefined, marketIDs: Array<Address>|null|undefined, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err: Error|null, result?: UIMarketsInfo) => void): void {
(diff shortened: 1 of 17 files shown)


epheph:
If you think this is a good idea, i'll change the other getters to grab the fully joined data to allow sorting/filtering on joined fields .

tinybike:
Agree that we should avoid duplicated join logic everywhere for creation time!  I do think that if we have functions for "get market data joined with market state" or "get market data joined with blocks" those should be _separate_ (composable) getters, i.e. you shouldn't have to join with more tables than you need data from!

epheph:
Sure, that makes sense.

src/server/getters/get-markets-info.ts
```diff
@@ -6,7 +6,7 @@ import { reshapeOutcomesRowToUIOutcomeInfo, reshapeMarketsRowToUIMarketInfo, get
 import { sortDirection } from "../../utils/sort-direction";
 
 export function getMarketsInfo(db: Knex, universe: Address|null|undefined, marketIDs: Array<Address>|null|undefined, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err: Error|null, result?: UIMarketsInfo) => void): void {
-  let marketsQuery: Knex.QueryBuilder = getMarketsWithReportingState(db).select("blocks.timestamp as creationTime").leftJoin("blocks", "markets.creationBlockNumber", "blocks.blockNumber").orderBy("creationTime");
+  let marketsQuery: Knex.QueryBuilder = getMarketsWithReportingState(db).orderBy("creationTime");
```
tinybike: One thing I just noticed here: `getMarketsInfo` should be using the `sortBy`, `isSortDescending`, `limit`, and `offset` parameters (like the other getters do), rather than just sorting by creation time.

tinybike: This doesn't need to be fixed as part of this issue though, it is an unrelated thing!

epheph:
I might try to abstract that, too, that's a lot of repeated code. i'll throw that into this PR, seems related-ish

tinybike:
That's a good point, that is a fair amount of boilerplate that goes into every getter function.  Made a new issue for the sortBy etc. stuff (4444 on CH) but if you want to address it as part of this PR, I'm all for that :sunglasses:

epheph:
Do you have a preference for how to do it? Could wrap all the modifier parameters into a dictionary, so it's easy to pass them around to a helper function that modifies a QueryBuilder? Or construct a modifier function in the dispatch that gets passed in? Still getting the hang of how Javascript folk approach design. Could also just pass each of the 4 args in to the helper function

epheph:
@pgebheim so i'm leaving incorrect sort behavior in getMarketsInfo (with incorrect tests that pass). to be addressed in a separate PR. I looked into some other solutions, like generating a function to do the modifiers, and rolling all values up into an object for cleaner function signatures, but it ended up either requiring more complexity in the tests or another layer between the dispatch and the getter

PR #41, opened 2017-10-30
support storing and checking network-id

epheph:


MicahZoltu:
There is talk of network ID being deprecated and removed from the RPC.  I recommend understanding the future of network ID before depending on it.

epheph:
@MicahZoltu do you have more details? I found EIP 611, but that seems like it's just renaming the endpoint used to fetch the same value?

MicahZoltu:
That is the issue I am thinking of, I haven't really followed it.  Just know that it exists.  If it really is just rename of the endpoint then you all probably won't run into any trouble.  I was under the impression that they were going to get rid of the functionality and replace it with new functionality for identifying what network you are participating in.

epheph:
From the EIP: 

> Introduce eth_networkId that retains the current functionality of net_version, which could be used by Mist for example to figure out which network an underlying node it running on.

I wasn't aware of that EIP, thanks for the heads up, we'll need to make sure this endpoint migrates when the new one is available.

@pgebheim this PR is ready for review

PR #43, opened 2017-10-31
Return marketIDs sorted by requested order

epheph:
@pgebheim 

How do you suppose we should handle asking for non-existent markets? I'm currently returning null for that array item, like:
```
...
          "id": 7,
          "outstandingShares": 100,
          "price": 0.125
        }
      ]
    },
    null,
    {
      "id": "0x0000000000000000000000000000000000000002",
      "universe": "0x000000000000000000000000000000000000000b",
      "type": "binary",
      "numOutcomes": 2,
...
```

easy enough to change. Perhaps @stephensprinkle do you have an opinion? if a marketID is requested and it does not exist, would you like to see it omitted or kept in (such that you can rely on previous marketID indexes matching in the response?))


epheph:
https://app.clubhouse.io/augur/story/4444/getmarketsinfo-should-sort-limit-using-its-parameters

src/server/getters/get-markets-info.ts
```diff
@@ -1,23 +1,12 @@
...
-  if (marketIDs != null) {
-    query = query.whereIn("markets.marketID", marketIDs);
-  }
-  query = queryModifier(query, "volume", "desc", sortBy, isSortDescending, limit, offset);
-  console.log(query.toSQL());
+export function getMarketsInfo(db: Knex, marketIDs: Array<Address>|null|undefined, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err: Error|null, result?: UIMarketsInfo) => void): void {
+  let query: Knex.QueryBuilder = getMarketsWithReportingState(db);
+  if (marketIDs == null || !marketIDs.length) return callback(new Error("must include marketIDs parameter"));
+  query = query.whereIn("markets.marketID", marketIDs);
   query.asCallback((err: Error|null, marketsRows?: Array<MarketsRowWithCreationTime>): void => {
```
pgebheim: Shouldn't we just also join against the outcomes table instead of doing queries for each market as well?

This is the beginning of an N+1 query problem.

src/server/getters/get-markets-info.ts
```diff
@@ -25,7 +15,8 @@ export function getMarketsInfo(db: Knex, universe: Address|null|undefined, marke
       db("outcomes").where("marketID", marketsRow.marketID).asCallback((err: Error|null, outcomesRows?: Array<OutcomesRow>): void => {
         if (err) return nextMarketsRow(err);
         const outcomesInfo: Array<UIOutcomeInfo> = outcomesRows!.map((outcomesRow: OutcomesRow): UIOutcomeInfo => reshapeOutcomesRowToUIOutcomeInfo(outcomesRow));
-        marketsInfo.push(reshapeMarketsRowToUIMarketInfo(marketsRow, outcomesInfo));
+        const marketsIndex = marketIDs.indexOf(marketsRow.marketID);
+        marketsInfo[marketsIndex] = reshapeMarketsRowToUIMarketInfo(marketsRow, outcomesInfo);
```
epheph: Could have also done a single pass re-map in the callback below.

src/server/getters/get-markets-info.ts
epheph: The data in outcomes is structured:
```
          outcomes: [{
            id: 0,
            outstandingShares: 1000,
            price: 0.5,
          }, {
            id: 1,
            outstandingShares: 1000,
            price: 0.5,
          }],
```

epheph: When i've seen this implemented before, it's returning the exact same left-side over and over, and it is up to the client to accumulate. Unless you have a better pattern/toolset to deal with that?

pgebheim: It's always up to our client code to structure the data coming back from the DB.

In this case, you can use _.groupBy and _.map to aggregate the outcomes on a market.

The other options is to break this into two queries:

get markets IN list
get outcomes with marketid IN list

and then merge them in your getter logic.

epheph:
@pgebheim one other point to discuss is naming. I called the new `universe` filter endpoint `get-markets` . I figure we could, eventually, roll the other endpoints in there. awaiting-reporting just == getMarkets filtered by reportingState. In fact, both of those awaiting would be rolled into just a single filter within `get-markets`

epheph:
For posterity, in a slack conversation with @stephensprinkle , he requested that a request for markets:
```
1,missing,2
```
returns resposne
```
1,null,2
```

PR #44, opened 2017-11-01
first pass at using augur.js reporting state constants

epheph:
Changes reportingState in DB and code to a string, implements their storage/removal market-created and designated-report-submitted. More to support, but wanted to get feedback on the basic idea.


epheph:
One thought I had was that there is no specific blockchain event that moves a market from `PRE_REPORTING` to `DESIGNATED_REPORTING`, it's just that the endtime of the market comes. I figure we can check all markets against the current block time that are in a time-dependent state on new block? Seems best to deal with it in processing than at query time.

epheph:
#46 covers these changes now

PR #46, opened 2017-11-02
staked tokens normalization

epheph:


test/blockchain/log-processors/market-created.js
```diff
@@ -123,10 +128,11 @@ describe("blockchain/log-processors/market-created", () => {
             tag2: "TEST_TAG_2",
             volume: 0,
             sharesOutstanding: 0,
+            reportingState: "PRE_REPORTING",
```
pgebheim: Can we have these reporting states recorded as a typescript Enum so that we get some value checking in the compiler?

src/server/getters/get-reporting-history.ts
```diff
@@ -46,7 +48,7 @@ export function getReportingHistory(db: Knex, reporter: Address, marketID: Addre
         reportingWindow: row.reportingWindow,
         payoutNumerators,
         amountStaked: row.amountStaked,
-        reportingToken: row.reportingToken,
+        stakedToken: row.stakedToken,
```
pgebheim: `stakeToken` not `stakedToken`

epheph: renamed

epheph:
@pgebheim implemented.

A little gross putting the `enum` in `types.ts`, but it is typescript-specific . if you have a better place, i'm all for it. maybe `database.ts`? But it's going to be used in blockchain processing as well

src/types.ts
```diff
@@ -131,10 +146,11 @@ export interface MarketsRow {
   tag2: string|null;
   volume: string|number;
   sharesOutstanding: string|number;
+  marketStateID: number;
   reportingWindow: Address;
   endTime: number;
   finalizationTime?: number|null;
-  reportingState?: number|null;
+  reportingState?: string|null;
```
pgebheim: Should use the enum type here: `reportingState?: REPORTING_STATE | null`

src/types.ts
```diff
@@ -1,6 +1,21 @@
 import Augur from "augur.js";
 import * as Knex from "knex";
 
+export enum REPORTING_STATE {
```
pgebheim: Looks like the TS convention for Enums is camel case (because they're a user-defined type, like a class), so update to `ReportingState` and then use that type where you would have used `string`

src/types.ts
epheph: Totally

PR #50, opened 2017-11-09
actually send the error to the websocket client, instead of just logging it

epheph:
Aside from being correct, should help a lot for debugging

/cc @stephensprinkle


epheph:
What a response looks like now (instead of hanging):
```
$ node scripts/test-client.js
{
  "id": 7,
  "jsonrpc": "2.0",
  "error": {
    "code": -32602,
    "message": "Must provide reference to universe, specify universe, marketID, reportingWindow",
    "data": false
  }
}
```

PR #54, opened 2017-11-10
Expose getMarkets to RPC

pgebheim:


epheph:
totally...

epheph:
I'd like to get a full suite of test-client.js payloads, since unit tests dont' actually test the json dispatch and would have caught this.

PR #57, opened 2017-11-10
add script to change the mock user hash so UI can test.

bthaile:
added simple script so UI team can change mock user to their geth user and test UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "lodash.omit": "4.5.0",
     "pg": "7.3.0",
     "postinstall-build": "5.0.1",
+    "replace-in-file": "^3.0.0",
     "sqlite3": "3.1.8",
     "uuid": "3.1.0",
     "ws": "3.2.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
     "lodash.omit": "4.5.0",
     "pg": "7.3.0",
     "postinstall-build": "5.0.1",
+    "replace-in-file": "^3.0.0",
```
epheph: should we make this a devDependency?

scripts/replace-user.js
```diff
@@ -0,0 +1,26 @@
...
+"use strict";
+const replace = require('replace-in-file');
+/*
+	usage: node replace-user <hash value>
+	run in augur-node root dir
+*/
+
+if (process.argv.length < 3) throw new Error('Pass in user hash as arguments');
+
+process.argv.slice(2).map(function (user_hash) {
```
epheph: Also, the structure of this program is to iterate on all arguments, and i don't think it makes sense to run through this map twice. If you want to just leave a lot of structure intact, you could just 

`slice(2,3)` and make sure the process.argv.length == 3

epheph:
Looks great. already approved earlier.

PR #58, opened 2017-11-10
get unfinalized stake tokens

epheph:
@stephensprinkle wasn't sure what details this endpoint was supposed to return, so please review for sanity.

The top-level object is the stakeToken address, it is not indexed by the reportingState.


tinybike:
This endpoint (like all the others) is supposed to return whatever details @stephensprinkle and team need for use in the UI.  This looks fine to me, provided it's supplying the UI with the data it needs.

PR #60, opened 2017-11-12
simple fix for open orders param filtering

epheph:
So we really need to resolve the issue of typescript types being way too loose and not actually ensuring that something that is.

`Address|null`

is not actually `undefined`

This is the third time it has burned us.


epheph:
I'd create a test for it, but since it is coming from javascript, and not employing any type safety at all, i'm not sure it would catch this specific issue.

tinybike:
Looks good!

PR #62, opened 2017-11-14
return better error on missing required parameter

epheph:
@bthaile


bthaile:
Do you need the `query = query` to build the query? Looks like you took it out.

epheph:
you don't need it, the query builder simply return this to keep acting on itself.

looking into test failure

PR #64, opened 2017-11-14
Fixed Categorical Seed Market Descriptions.

JohnDanz:
Fixes some Seed data around categorical markets and their descriptions. The UI is expecting them to have the outcome labels included and without the special characters/labels it ends up completely stripping out the description of the market in the UI.


epheph:
Will this format be important for real market titles taken off the event stream?

Tests will need to be updated, let me know if you'd like me to do it.

PR #67, opened 2017-11-15
remove better/worse OrderID

epheph:
Discussed with @nuevoalex and @stephensprinkle , this field is an implementation detail of solidity and not necessary, since we have a database.

Additionally, I'm concerned about the validity of data that is sourced both an event and an rpc request, since we have no guarantees about the state of the rpc chain matches the state when the event was validated


tinybike:
It's fine to remove this from the table, but these IDs _do_ need to be passed to the UI.  Better/worse order ID are needed by the UI when it submits trades for on-chain order matching to work correctly.

PR #85, opened 2017-11-22
Change process-block to accept actual block object, not just number

epheph:
/cc @MicahZoltu

### DIFF
--- a/src/blockchain/process-block.ts
+++ b/src/blockchain/process-block.ts
@@ -1,39 +1,37 @@
 import Augur from "augur.js";
 import * as Knex from "knex";
 import { logError } from "../utils/log-error";
-import { Int256, BlocksRow } from "../types";
+import { Block, BlocksRow, Int256 } from "../types";
 
-export function processBlock(db: Knex, augur: Augur, blockNumberString: Int256): void {
-  augur.rpc.eth.getBlockByNumber([blockNumberString, false], (block: any): void => {
-    if (!block || block.error || !block.timestamp) return logError(new Error(JSON.stringify(block)));
-    const blockNumber = parseInt(blockNumberString, 16);
-    const blockHash = block.hash;
-    const timestamp = parseInt(block.timestamp, 16);
-    console.log("new block:", blockNumber, timestamp);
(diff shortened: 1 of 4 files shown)


src/blockchain/process-block.ts
```diff
@@ -1,39 +1,37 @@
 import Augur from "augur.js";
 import * as Knex from "knex";
 import { logError } from "../utils/log-error";
-import { Int256, BlocksRow } from "../types";
+import { Block, BlocksRow, Int256 } from "../types";
```
MicahZoltu: This actually works/compiles and you get type info for `Block`?  I don't see you exporting `Block` from `types.ts` below so this working surprises me...

src/blockchain/start-augur-listeners.ts
```diff
@@ -27,11 +27,11 @@ export function startAugurListeners(db: Knex, augur: Augur, callback: (blockNumb
       Transfer: makeLogListener(db, augur, "LegacyReputationToken", "Transfer"),
       Approval: makeLogListener(db, augur, "LegacyReputationToken", "Approval"),
     },
-  }, (blockNumber: Int256): void => {
+  }, (block: Block): void => {
```
MicahZoltu: FYI: If we ever get full type defenitions for `augur.js` you'll be able to change this to `block => { ... }` and type inference will figure out the rest.

epheph: Actually, we have defined the type for this, so i can technically remove it. Is there a best practice for when you do so? just, whenever possible?

epheph: https://github.com/AugurProject/augur.js/blob/91539115d4c32648df3cb77b65f1859fd456629f/index.d.ts#L169

MicahZoltu: I'm personally a big fan of letting type inference do its job.  However, I also am a big fan in running TS in ultra-strict mode which doesn't allow for implicit `any` types so you get a complaint when the type inferer _can't_ do its job.  Without strict mode, I would be hesitant to follow my usual pattern of "remove type annotations where possible" since it is too easy to fall back into "infer `any`" mode.

PR #94, opened 2017-11-27
first pass at downloading blocks when using getAllAugurLogs

epheph:
@tinybike thoughts about correct behavior when processing block fails? there were no existing callbacks in processBlock, so I didn't add it in this PR, but it's probably correct to add it.

I'm doing the whole thing as 1 pass at the beginning to ease de-duplication of blockNumbers without adding extra logic instead of the nested `eachSeries`

### DIFF
--- a/src/blockchain/download-augur-logs.ts
+++ b/src/blockchain/download-augur-logs.ts
@@ -4,11 +4,14 @@ import * as Knex from "knex";
 import { ErrorCallback, FormattedEventLog } from "../types";
 import { processLog } from "./process-logs";
 import { logProcessors } from "./log-processors";
+import { processBlockByNumber } from "./process-block";
 
 export function downloadAugurLogs(db: Knex, augur: Augur, fromBlock: number, toBlock: number, callback: ErrorCallback): void {
   console.log("Getting Augur logs from block " + fromBlock + " to block " + toBlock);
   augur.events.getAllAugurLogs({ fromBlock, toBlock }, (err?: string|object|null, allAugurLogs?: Array<FormattedEventLog>): void => {
     if (err) return callback(err instanceof Error ? err : new Error(JSON.stringify(err)));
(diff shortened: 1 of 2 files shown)


src/blockchain/download-augur-logs.ts
```diff
@@ -31,3 +34,13 @@ export function downloadAugurLogs(db: Knex, augur: Augur, fromBlock: number, toB
     ), callback);
   });
 }
+
+function downloadBlocksFromLogs(db: Knex, augur: Augur, allAugurLogs?: AugurLogs): void {
+  const blockNumbers = _.uniq(
```
tinybike: Unless Lodash has some fancy optimization to avoid this, this `blockNumbers` calculation will generate an array with an entry for each log, from which a smaller array w/ just the unique numbers is extracted.  Possibly more efficient to define `const blockNumbers = {}` and add a line inside the nested `eachSeries`:
```javascript
if (!blockNumbers[log.blockNumber]) blockNumbers[log.blockNumber] = true;
```
and then use `Object.keys(blockNumbers)` to loop over the numbers to pass to `processBlockByNumber`.

(Possibly a premature optimization, just jumped out at me a bit when reading.)

tinybike:
Adding a callback to processBlock sounds reasonable to me.  Right now, processBlock just prints to the console if there's an error (`logError`), which is fine as a default, but it'd be sensible to make the callback a parameter.

src/blockchain/download-augur-logs.ts
epheph: I was trying to avoid having unlinked data in the database, where we're inserting the log events before the blocks (and hence would need the callback) but maybe it isn't that important.

The way i've done it does feel a bit too pythonic to me, i'll change.

epheph:
@tinybike force pushed new update, significantly simplified with new structure

PR #96, opened 2017-11-28
Fix designated reporter processor, implement removal

epheph:
@tinybike this should be good to go, wrote the test for it (although haven't tested against a real blockchain process yet)


epheph:
closed in favor of #101

PR #99, opened 2017-11-28
outcomes seeds for all markets

epheph:


epheph:
Something is wrong with travis integration, this change passes

PR #100, opened 2017-11-28
Revert "Hack to fix missing fields for now"

epheph:
Reverts AugurProject/augur-node#98


epheph:
Merge this once we have a new contracts uploaded with canned data applied

epheph:
Fixed in another PR

PR #111, opened 2017-12-04
Update constraint names

tinybike:
`positive` --> `nonnegative`


tinybike:
Merging this because it is Serious Business

PR #114, opened 2017-12-05
4911 human readable shares

tinybike:
Convert to human-readable share quantities for the UI (and for augur-node's internal use).  Note: the matching changes in augur.js are in https://github.com/AugurProject/augur.js/pull/191


tinybike:
Implements suggested scheme in CH 4911; @joeykrug if you have time please glance over this and see if I'm messing anything up...

epheph:
👍

PR #115, opened 2017-12-05
Add Docker support to Augur Node

pgebheim:
Included is a Docker file which exposes the RPC port (9001) and a docker-compose file which can be used to build and name this image with a :dev tag for convenience.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,25 @@
-FROM node:8
+FROM node:8 as builder
 
-EXPOSE 9001
+WORKDIR /root/
+COPY config.json config.json
+COPY tsconfig.json tsconfig.json
+COPY package.json package.json
+COPY src src
 
-COPY config.json /root/augur-node/config.json
-COPY tsconfig.json /root/augur-node/tsconfig.json
-COPY package.json /root/augur-node/package.json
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:8 as builder
+
+WORKDIR /root/
```
MicahZoltu: style: I'm personally a fan of `/app` since `/root` can sometimes have stuff in it already from base image (I have seen `/root` map to `~`).

epheph:
Nice

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN npm run build
+
+COPY knexfile.js knexfile.js
+RUN npm run migrate
+
+FROM node:8
+EXPOSE 9001
+WORKDIR /root/
+COPY --from=builder /root/build build
+COPY --from=builder /root/node_modules node_modules
```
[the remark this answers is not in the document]
pgebheim: That's a good improvement, once we do that I can update the docker file.

Dockerfile
pgebheim: Agreed, I'll update.

PR #119, opened 2017-12-07
add emitters for a bunch of stuff, while i was doing TokensTransferred

epheph:
@stephensprinkle 

https://app.clubhouse.io/augur/story/5019/tokenstransferred-event-not-emitted-when-transfering-tokens


epheph:
@stephensprinkle I'm not sure what the UI is expecting here, but copying the pattern from other processors, we basically send the db row that we are about to insert on log addition, and the raw event log on event removal (which will contain the `removed: true`

epheph:
@JohnDanz look good?

PR #120, opened 2017-12-08
reportingWindow table and log processor

epheph:


epheph:
thx for the reporting window event @nuevoalex

PR #124, opened 2017-12-12
Added reinstall and sqlite repl scripts

tinybike:
Just a couple quality-of-life scripts I run locally, figured it might be worth adding them to package.json!


epheph:
Removing test-client?

PR #129, opened 2017-12-14
use env OR defaults, not both

epheph:
@MicahZoltu per our discussion


epheph:
i'll write a type for the endpoints

PR #131, opened 2017-12-15
first pass at binding to both wss(port 9002) and ws(port 9001)

epheph:


src/server/run-websocket-server.ts
```diff
@@ -2,68 +2,86 @@ import * as WebSocket from "ws";
...
-      }
+  const websocketServers: Array<WebSocket.Server> = [];
 
-      try {
-        if (message.method === "subscribe") {
-          const eventName: string = message.params.shift();
+  if ( webSocketConfigs.wss != null ) {
+    console.log("Starting websocket secure server on port", webSocketConfigs.wss.port);
+    const server = https.createServer({
+      cert: fs.readFileSync(webSocketConfigs.wss.certificateFile),
```
epheph: Reasonable way to read a file here? Mostly what i've read says that it's OK to do sync things during setup when you can't proceed until it is done

epheph:
Still need env override functionality

src/server/run-websocket-server.ts
pgebheim: yeah thats fine, can't start up without it if its specified.

PR #141, opened 2017-12-27
Deal with variably padded addresses

kennsippell:
https://app.clubhouse.io/augur/story/5104/deal-with-variably-padded-addresses

### DIFF
--- a/src/server/address-format-reviver.ts
+++ b/src/server/address-format-reviver.ts
@@ -1,0 +1,31 @@
+/*
+A reviver is a function that prescribes how the value originally produced by JSON.parse is transformed, before being returned.
+*/
+
+import { formatEthereumAddress } from "speedomatic";
+
+const inputsExpectedAsAddress : string[] = [
+  "account",
+  "author",
+  "creator",
+  "filler",
+  "designatedReporter",
+  "marketCreator",
+  "marketID",
+  "orderCreator",
+  "owner",
+  "reportingWindow",
+  "shareToken",
+  "stakeToken",
+  "token",
+  "universe",
+];
+
+export function addressFormatReviver (key : any, value : any) {
+  console.log('key', key);
+  if (inputsExpectedAsAddress.some(x => x === key)) {
+    return formatEthereumAddress(value);
+  }
+
+  return value;
+};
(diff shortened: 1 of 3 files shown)


src/server/address-format-reviver.ts
```diff
@@ -0,0 +1,31 @@
+/*
+A reviver is a function that prescribes how the value originally produced by JSON.parse is transformed, before being returned.
+*/
+
+import { formatEthereumAddress } from "speedomatic";
+
+export const inputsExpectedAsAddress: Array<string> = [
```
pgebheim: Use the Object-as-set trick to turn this into O(1) Lookups.

src/server/address-format-reviver.ts
```diff
@@ -0,0 +1,32 @@
...
+  [key: string]: boolean;
+}
+
+export const inputsExpectedAsAddress: StringToBoolMap = {
+  account: true,
+  category: true,
+  creator: true,
+  designatedReporter: true,
+  marketID: true,
+  marketIDs: true,
```
epheph: I was going to comment that this won't work, but after reading formatEthereumAddress and seeing it works with arrays, looks good.

src/server/address-format-reviver.ts
```diff
@@ -0,0 +1,32 @@
...
+  reporter: true,
+  reportingWindow: true,
+  result: true,
```
epheph: result? I don't know if this is necessary as a parameter to a getter

src/server/address-format-reviver.ts
kennsippell: ya. got a unit test for that case also

src/server/run-websocket-server.ts
```diff
@@ -36,7 +37,7 @@ export function runWebsocketServer(db: Knex, webSocketConfigs: WebSocketConfigs)
       websocket.on("message", (data: WebSocket.Data): void => {
         let message: any;
         try {
-          message = JSON.parse(data as string);
+          message = JSON.parse(data as string, addressFormatReviver);
```
epheph: nice dealing with it in the parser

src/server/address-format-reviver.ts
kennsippell: ic -- it is the name of the variable in the "callback" from the definitions. removing

PR #143, opened 2017-12-28
Return timestamp in result from get-markets-created-by-user

kennsippell:
https://app.clubhouse.io/augur/story/4855/return-timestamp-in-result-from-get-markets-created-by-user-and-get-reporting-history

### DIFF
--- a/scripts/payloads/getMarketsCreatedByUser.json
+++ b/scripts/payloads/getMarketsCreatedByUser.json
@@ -1,0 +1,13 @@
+{
+  "id": 5,
+  "jsonrpc": "2.0",
+  "method": "getMarketsCreatedByUser",
+  "params": {
+    "universe": "0x000000000000000000000000000000000000000b",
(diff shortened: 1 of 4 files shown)


kennsippell:
Note this work item also calls for creationTime in get-reporting-history, but this appears to already be in place.

epheph:
Looks great.

@bthaile @stephensprinkle this changes the format of the response, from an array to an object, so we can include timestamp. this was implemented as 
```
[{ marketID, timestamp}]
```

but would probably be just as easy to write as
```{ marketID: timestamp }```

if that's what the UI needs?

thx @kennsippell

kennsippell:
it's actually ```[{ marketID, creationTime }]```

epheph:
discussion in #ui, we should use our existing strategy, "you get marketIDs from various APIs, then feed them into getMarketsInfo to resolve more details" and not bother marketID-returning APIs with returning specific data

PR #147, opened 2018-01-01
bump sqlite3 version to one with available binary for node latest (9.3.0)

stephensprinkle:
Referenced GitHub issue: https://github.com/mapbox/node-sqlite3/issues/918

Confirmed this as working in both Mac + Linux environments against node latests + lts.


epheph:
👍

PR #150, opened 2018-01-09
implement configurable websocket ping

epheph:


epheph:
augur.js websocket responds appropriately with pong (although due to websocket xor mask'ing, it's not immediately obvious in a tcpdump)

PR #152, opened 2018-01-09
Add database status endpoint

pgebheim:


src/server/run-server.ts
```diff
@@ -40,5 +40,14 @@ export function runServer(db: Knex, augur: Augur): RunServerResult {
     }
   });
 
+  app.get("/status/database", (req, res) => {
+    const maxPendingTransactions: number = (typeof req.query.max_pending_transactions === "undefined") ? 1 : parseInt(req.query.max_pending_transactions, 10);
+    if (isNaN(maxPendingTransactions)) {
+      res.status(400).send({error: "Bad value for max_pending_transactions, must be an integer in base 10"});
```
epheph: pedantic, but 500?

epheph: can we also include: `status: down` so one could just search for that string?

epheph:
Looks great

PR #159, opened 2018-01-13
index endTime for market finalized check, happens every block

epheph:


epheph:
I should convert all these migrations to knex native ones

PR #161, opened 2018-01-18
Wrong comparison in advanceTime

epheph:
@bthaile


src/blockchain/sync-augur-node-with-blockchain.ts
```diff
@@ -15,6 +15,7 @@ interface NetworkIDRow {
 function getNetworkID(db: Knex, augur: Augur, callback: (err: Error|null, networkID: string|null) => void) {
   const networkID: string = augur.rpc.getNetworkID();
   db.select("networkID").from("network_id").limit(1).asCallback( (err: Error|null, rows: Array<NetworkIDRow>): void => {
+    if (err) return callback(err, null);
```
epheph: not related, but threw it in there anyway

PR #164, opened 2018-01-23
Implement crowdsourcer created, contributed, completed

epheph:


src/blockchain/log-processors/crowdsourcer.ts
```diff
@@ -2,42 +2,54 @@ import Augur from "augur.js";
...
+      blockNumber: log.blockNumber,
+      transactionHash: log.transactionHash,
+      logIndex: log.logIndex,
+      crowdsourcerID: log.disputeCrowdsourcer,
+      marketID: log.market,
+      size: log.size,
+      payoutID,
+    };
+    db.transacting(trx).insert(crowdsourcerToInsert).into("crowdsourcers").returning("crowdsourcerID").asCallback(callback);
+    augurEmitter.emit("DisputeCrowdsourcerCreated", log);
```
pgebheim: Do we want these Emitted before the DB write completes?

I think it works, but worth a double check there's no funny race condition. Main issue I see is if that insert fails we've emitted an event that isn't look-up-able in the DB.

src/blockchain/log-processors/database.ts
```diff
@@ -10,3 +10,25 @@ export function updateMarketState(db: Knex, marketID: Address, trx: Knex.Transac
...
+  const payoutRow: { [index: string]: string|number|boolean|null } = {
+    marketID,
+    isInvalid: invalid,
+  };
+  payoutNumerators.forEach((value: number, i: number): void => {
+    payoutRow["payout" + i] = value;
+  });
+  db.transacting(trx).select("payoutID").from("payouts").where(payoutRow).first().asCallback( (err: Error|null, payoutID?: number|null): void => {
+    if (err) return callback(err);
+    if (payoutID != null) {
```
pgebheim: Is this to stop duplicate insertion by PayoutID?

Since we're writing to the DB Serially this should work, but it is an anti-pattern as its likely to create issues if we ever move to more than one writer at a time (e.g. when we move to Postgres or another DB that supports multiple connections).

It's better to actually create the unique constraint in the DB, do the write and catch the error, so that the synchronization is done at the DB layer, avoiding the race.

epheph: I agree, i was having difficulty coming up with a pattern that both worked for psql AND sqlite AND returned the insert id. I'll fix it to use replace/on duplicate and select after perhaps.

src/blockchain/log-processors/crowdsourcer.ts
epheph: You're right, I was being lazy, fixing

src/blockchain/log-processors/database.ts
epheph: In this specific case, there will not be multiple connections to introduce a race, even in postgres, because of the way we've linearized the processing of logs, but i still agree about fixing it

PR #169, opened 2018-01-26
initial report size

epheph:
After reading whitepaper, i realized the initial report size is both dynamic and important (to determine crowdsource bond sizes). When this event fires is our only opportunity to grab that value.


epheph:
thx for the help @nuevoalex

PR #174, opened 2018-01-31
fee window advance

epheph:
Now that we expose fee windows, it depends on logic to advance the time forward (since blocktime exceeding endTime is how we know the window has passed)

### DIFF
--- a/src/blockchain/process-block.ts
+++ b/src/blockchain/process-block.ts
@@ -83,10 +83,10 @@ function _processBlockRemoval(db: Knex, block: Block, callback: ErrorCallback): 
 }
 
 function advanceTime(db: Knex, augur: Augur, trx: Knex.Transaction, blockNumber: number, timestamp: number, callback: AsyncCallback) {
-  parallel( {
-    advanceMarketReachingEndTime: (next: AsyncCallback) => advanceMarketReachingEndTime(db, augur, trx, blockNumber, timestamp, next),
-    advanceFeeWindowActive: (next: AsyncCallback) => advanceFeeWindowActive(db, trx, blockNumber, timestamp, next),
-  }, callback);
+  parallel( [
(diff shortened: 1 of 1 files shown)


src/blockchain/process-block.ts
```diff
@@ -79,18 +84,37 @@ function _processBlockRemoval(db: Knex, block: Block, callback: ErrorCallback):
...
 
 function advanceMarketReachingEndTime(db: Knex, augur: Augur, trx: Knex.Transaction, blockNumber: number, timestamp: number, callback: AsyncCallback) {
   const designatedDisputeQuery = db("markets").transacting(trx).select("markets.marketID").join("market_state", "market_state.marketStateID", "markets.marketStateID");
   designatedDisputeQuery.where("reportingState", augur.constants.REPORTING_STATE.PRE_REPORTING).where("endTime", "<", timestamp);
-  designatedDisputeQuery.asCallback( (err: Error|null, designatedDisputeMarketIDs: Array<any> ) => {
+  designatedDisputeQuery.asCallback((err: Error|null, designatedDisputeMarketIDs: Array<any>) => {
     if (err) return callback(err);
     each(designatedDisputeMarketIDs, (marketIDRow, nextMarketID: ErrorCallback) => {
-      updateMarketState(db, marketIDRow.marketID, trx, blockNumber, augur.constants.REPORTING_STATE.DESIGNATED_REPORTING, nextMarketID);
+      updateMarketState(db, marketIDRow.marketID, trx, blockNumber, augur.constants.REPORTING_STATE.DESIGNATED_REPORTING, (err: Error|null) => {
```
pgebheim: Can you do this in one query with the list of market ids in an IN?

epheph: updating market state is more complex than an update, it manages a one-to-many mapping, inserting a new row into market_state and linking to that row (so we both have a history of when these changes happened and for easy undo by re-linking to the previous row)

PR #178, opened 2018-02-02
gotta go fast

epheph:
so much faster


epheph:
fixed up a few types and object usages while i was at it. we're jumping from 1.49m to 1.7m to scan at start

PR #182, opened 2018-02-05
schema and processing changes for upcoming get-dispute-info endpoint

epheph:


epheph:
This adds:
initial_reports.amountStaked,
crowdsourcers.amountStaked,
crowdsourcers.feeWindow (pulled from current active feeWindow, not emitted from contract)

PR #183, opened 2018-02-06
get dispute info endpoint

epheph:
There are two items remaining:
1.) Dispute round # (or null)
2.) Passing in reporter for how much that reporter has staked on each payout per market

/cc @stephensprinkle 

```
$ node test-client.js payloads/getDisputeInfo.json 

payloads/getDisputeInfo.json
{
  "id": 61,
  "jsonrpc": "2.0",
  "method": "getDisputeInfo",
  "params": {
    "marketIDs": ["0x0000000000000000000000000000000000000011","0x0000000000000000000000000000000000000211"]
  }
}

{
  "id": 61,
  "result": [
    {
      "marketID": "0x0000000000000000000000000000000000000011",
      "stakes": [
        {
          "isInvalid": false,
          "payout": [
            0,
            2
          ],
          "totalStaked": 20000,
          "size": 20000,
          "amountStaked": 20000
        },
        {
          "isInvalid": true,
          "payout": [
            1,
            1
          ],
          "totalStaked": 20000,
          "size": 40000,
          "amountStaked": 20000
        }
      ]
    },
    {
      "marketID": "0x0000000000000000000000000000000000000211",
      "stakes": [
        {
          "isInvalid": false,
          "payout": [
            0,
            10000
          ],
          "totalStaked": 122,
          "size": 204,
          "amountStaked": 20
        }
      ]
    }
  ],
  "jsonrpc": "2.0"
}
```


src/types.ts
```diff
@@ -238,6 +238,17 @@ export interface UIDisputeTokens {
...
+export interface StakeDetails extends Payout {
+  totalStaked: number;
+  size: number;
+  amountStaked: number;
+}
+
+export interface UIStakeInfo {
+  marketID: Address;
+  stakes: Array<StakeDetails>;
+}
```
bthaile: Is there a mapping to Market Outcome and payout? In UI there is just outcomes and TODOs for mapping market payoutNumerators to outcomes.

src/types.ts
```diff
@@ -238,6 +238,17 @@ export interface UIDisputeTokens {
   [stakeToken: string]: UIDisputeTokenInfo;
 }
 
+export interface StakeDetails extends Payout {
+  totalStaked: number;
+  size: number;
+  amountStaked: number;
+}
```
bthaile: definitely what the UI needs per market dispute, so an array of these works.

src/server/getters/get-dispute-info.ts
```diff
@@ -0,0 +1,76 @@
...
+  initialReport: Array<StakeRow>;
+}
+
+interface StakeRow extends Payout {
+  marketID: Address;
+  amountStaked: number;
+  payoutID: number;
+  size?: number;
+  completed: number|null;
+}
```
bthaile: can number handle the size of attoREP? could be 10x-18 long, right.

epheph: you are correct, it is not the correct numeric container. we have a different CH for going through fixing all these, but i'll go ahead and do this one right now

src/types.ts
epheph: under stakes, there will be payoutNumerators, so i think you'd have to see which one was >0 and which one was == 0?

src/server/getters/get-dispute-info.ts
epheph: Actually, it looks like this might be a bit large to do correctly in the same PR, it should be a different PR since we need to change the getter mechanisms from the DB (and make it work with both postgres and sqlite), or else the number could get squashed on its way out from DB.

I could just half-ass it right now and bignumber/toFixed() it on the way out so it doesn't change for you when we fix it for realz?

bthaile: is that storing as hex like straight from contract log? either way you da man.

src/types.ts
bthaile: discussion in #dev discord

bthaile:
looks good enough to get us down the road. looks like known super big number issue.

src/server/getters/get-dispute-info.ts
epheph: no we will serve numbers as base-10, strings

PR #184, opened 2018-02-08
ensure we process blocks in correct order, in between logs by block

epheph:


epheph:
Going to do different, @pgebheim and i discussed having only blocks be in the queue, and it is the processing of blocks that brings a block's logs along for the ride

PR #187, opened 2018-02-09
Updated designated reporting label

tinybike:
Also added a market ID and outcome lookup to order-canceled log processor and attach it to the emitter.


tinybike:
No, there's just an open PR in the UI repo that needs these fields!

PR #189, opened 2018-02-09
timestamp override

epheph:


epheph:
@bthaile this PR makes block numbers inside augur-node (and anything that links to them) overridden by a TimestampSet blockchain event

PR #193, opened 2018-02-14
just updating get-markets for now. Some other methods could be deprecated for the most part

nuevoalex:
This will allow the UI to filter all market requests by creator for the bug bounty. May provide useful to have one big generic method for other reasons too.


epheph:
bug bounty?

nuevoalex:
In the initial release we want to filter out any market not made by a sanctioned dev account. The market(s) created by that account is the bug bounty market, where the question will be something to the effect of "Will this market be hacked".

PR #194, opened 2018-02-14
Add pre-push hook to run lint and unit tests

justinbarry:


justinbarry:
@epheph Ironically, it failed here. Any clue why?

PR #197, opened 2018-02-15
log.market needs to be log.marketId

bthaile:
update log for create market to return name that reps the value.


bthaile:
incorrect location and need a more robust property translaction solution

PR #198, opened 2018-02-16
move markets into open_reporting if designated reporter doesn't show up.

epheph:
requires new constants in augur.js


epheph:
@justinbarry this change moves markets into OPEN_REPORTING state, inspecting the condition every new block

justinbarry:
Awesome. Thanks!

PR #201, opened 2018-02-17
Get initial reporters

nuevoalex:
This also includes some updates to the getter for dispute tokens as well as support for generic token balance tracking with the balances table.


epheph:
This PR has introduced a problem and augur-node is currently unstable. Not a huge deal but something we'll have to fix. We now get

```
insert into "transfers" ("blockNumber", "logIndex", "recipient", "sender", "token", "transactionHash", "value") values (1703699, 0, '0x113b462d14c542d208f5262d82e2eafd7cffd88a', NULL, NULL, '0x23def7f34f3eeed54b0475842975e9cceede4fed7530f52f6ea1948ff5d2dbc5', NULL) - SQLITE_CONSTRAINT: NOT NULL constraint failed: transfers.token
```

The problem is that the "Mint" processor tries to read the token off the log, but it's not present. the problem is that there is an Augur:

 `event TokensMinted(address indexed universe, address indexed token, address indexed target, uint256 amount);`

and the token itself has a:

`event Mint(address indexed target, uint256 value);`

Do _ALL_ mints we care about go through TokensMinted?

PR #205, opened 2018-02-20
Fee windows

nuevoalex:
Getter for participation tokens / fee windows by user. Added a token supply table for ease of use regarding the expected fees. Could do similar things for other stake tokens now too.


src/utils/get-current-time.ts
```diff
@@ -0,0 +1,5 @@
+import { getOverrideTimestamp } from "../blockchain/process-block";
+
+export function getCurrentTime(): number {
+  return  getOverrideTimestamp() || Date.now() / 1000;
```
epheph: Bringing system time to blockchain time seems like it could cause unforeseen problems? I wonder if this would be better served getting the head block time?

src/blockchain/log-processors/token/mint.ts
```diff
@@ -15,13 +18,19 @@ export function processMintLog(db: Knex, augur: Augur, trx: Knex.Transaction, lo
     blockNumber:     log.blockNumber,
   }).into("transfers").asCallback((err: Error|null): void => {
     if (err) return callback(err);
-    increaseTokenBalance(db, augur, trx, log.token || log.address, log.target, Number(log.amount || log.value), callback);
+    parallel([
+      (next: AsyncCallback): void => increaseTokenSupply(db, augur, trx, log.token || log.address, Number(log.amount || log.value), next),
+      (next: AsyncCallback): void => increaseTokenBalance(db, augur, trx, log.token || log.address, log.target, Number(log.amount || log.value), next),
```
epheph: `log.amount || log.value` is referenced 3 times, it's probably worth a variable. Probably all those solidity stack depth limitation issues you had to deal with ;)

src/utils/get-current-time.ts
nuevoalex: Yea that's a good point. Will change.

src/blockchain/log-processors/token/mint.ts
nuevoalex: Haha, yea will do.

src/blockchain/process-block.ts
```diff
@@ -53,6 +58,7 @@ function _processBlock(db: Knex, augur: Augur, block: Block, callback: ErrorCall
   if (!block || !block.timestamp) return logError(new Error(JSON.stringify(block)));
   const blockNumber = parseInt(block.number, 16);
   const blockHash = block.hash;
+  blockHeadTimestamp = block.timestamp;
```
epheph: block.timestamp is a hex string here.

PR #206, opened 2018-02-20
extract the getPosition/updatePosition method

epheph:


epheph:
just extracts the getTickSize-> getPositionInMarket->upsertPositionInMarket pattern

epheph:
merging to avoid large conflict in block seq pr

PR #211, opened 2018-02-22
Stake token fixes

nuevoalex:
Some additions to log listeners as well as some tweaks to the initial reporters getter

### DIFF
--- a/src/server/getters/get-initial-reporters.ts
+++ b/src/server/getters/get-initial-reporters.ts
@@ -1,14 +1,23 @@
 import * as Knex from "knex";
+import Augur from "augur.js";
 import { Address } from "../../types";
-import { InitialReportersRow } from "../../types";
+import { queryModifier } from "./database";
+import { InitialReportersRow, UIInitialReporters } from "../../types";
 
-export function getInitialReporters(db: Knex, reporter: Address, redeemed: boolean|null|undefined, callback: (err: Error|null, result?: any) => void): void {
-  let query = db.select(["marketId", "reporter", "amountStaked", "initialReporter", "redeemed"]).from("initial_reports").where({ reporter });
+export function getInitialReporters(db: Knex, augur: Augur, reporter: Address, redeemed: boolean|null|undefined, withRepBalance: false, callback: (err: Error|null, result?: any) => void): void {
+  let query = db.select(["marketID", "reporter", "amountStaked", "initialReporter", "redeemed", "isDesignatedReporter", "balances.balance AS repBalance"]).from("initial_reports")
+    .join("balances", function () {
+      this
+        .on("balances.token", augur.contracts.addresses[augur.rpc.getNetworkID()].ReputationToken)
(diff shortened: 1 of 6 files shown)


src/server/getters/get-initial-reporters.ts
```diff
@@ -1,14 +1,23 @@
 import * as Knex from "knex";
+import Augur from "augur.js";
 import { Address } from "../../types";
-import { InitialReportersRow } from "../../types";
+import { queryModifier } from "./database";
+import { InitialReportersRow, UIInitialReporters } from "../../types";
 
-export function getInitialReporters(db: Knex, reporter: Address, redeemed: boolean|null|undefined, callback: (err: Error|null, result?: any) => void): void {
-  let query = db.select(["marketId", "reporter", "amountStaked", "initialReporter", "redeemed"]).from("initial_reports").where({ reporter });
+export function getInitialReporters(db: Knex, augur: Augur, reporter: Address, redeemed: boolean|null|undefined, withRepBalance: false, callback: (err: Error|null, result?: any) => void): void {
```
adrake33: Why is the `withRepBalance` parameter set to false?  Shouldn't it look more like the `redeemed` parameter instead (i.e., `withRepBalance: boolean|null|undefined`)?

I'm also not sure I understand why the balance would be different than the amount staked. Can you please explain this?

src/types.ts
```diff
@@ -461,11 +461,12 @@ export interface FeeWindowRow {
...
 
 export interface InitialReportersRow {
-  initialReportId: number;
   marketId: Address;
   reporter: Address;
-  isDesignatedReporter: boolean;
   amountStaked: number;
+  initialReporter: number;
+  redeemed: boolean;
+  isDesignatedReporter: boolean;
```
adrake33: Doesn't `repBalance` also need to be added to this interface?

src/server/getters/get-initial-reporters.ts
nuevoalex: `withRepBalance` normally isn't needed as `redeemed` is sufficient. Even though this will be used elsewhere my initial use case is for the escape hatch, which is weird because even if the initial reporter is not redeemed it can potentially have already been emptied through the emergency withdraw method, so I need to conditionally get the ones that actually have a remaining balance

src/types.ts
nuevoalex: Yup!

src/server/getters/get-initial-reporters.ts
adrake33: Ok, so when `augur.reporting.getInitialReporters` is called from the augur.js side, will it accept `withRepBalance` as a parameter, or will it always pass in `false` for `withRepBalance` when calling this getter?

nuevoalex: Oh actually I may have typoed that looking again. I'm pretty sure you're correct haha, will address

PR #213, opened 2018-02-22
add creator to the log emit, and put the emit in the correct spot chronologically

epheph:


src/blockchain/log-processors/order-filled/index.ts
```diff
@@ -104,7 +103,10 @@ export function processOrderFilledLogRemoval(db: Knex, augur: Augur, log: Format
         if (err) return callback(err);
         db.from("trades").where({ marketId, outcome, orderId, blockNumber }).del().asCallback((err?: Error|null): void => {
           if (err) return callback(err);
-          updateOrdersAndPositions(db, augur, marketId, orderId, orderCreator, log.filler, numTicks, tickSize, callback);
+          updateOrdersAndPositions(db, augur, marketId, orderId, orderCreator, log.filler, numTicks, tickSize, (err?: Error|null) => {
+            augurEmitter.emit("OrderFilled", Object.assign({}, log, { creator: orderCreator }));
+            callback(err);
```
adrake33: Should this be `return callback(err);` like on the lines above?

epheph: Normally we'd have "if (err) return callback(err)". I guess it's up to you, would you rather have the event emitted, even if we have an internal issue? Maybe it is weird to say "hey, check out this new data" and they're all "what new data?"

epheph:
changed the line, more important to be consistent

epheph:
changed the line, more important to be consistent

PR #219, opened 2018-02-26
Added share adjustments on-the-way-out in getUserTradingPositions

tinybike:
Resolves [ch7134]


tinybike:
Not sure if this is the best way to do this or not, should be simple to reverse if we decide against it though.  I left the `numSharesAdjustedForUserIntention` machinery in as well, in case someone makes a very compelling case that we ought to use it for something.

tinybike:
I'm going ahead and merging this (b/c it's needed for https://github.com/AugurProject/augur/pull/1168 to not display positions that are utter nonsense, and because 1168 is in turn blocking everything else).  I do still want feedback on it though @epheph if you have a sec.

PR #220, opened 2018-02-26
fix up timestamp numeric handling, fix date comparison bug, and other cleanup

epheph:
It's a race, if there is data in flight for a json rpc and the client goes away before it is sent, we'll get this (easy to recreate with a setTimeout in a getter).


src/server/run-websocket-server.ts
```diff
@@ -69,6 +69,10 @@ export function runWebsocketServer(db: Knex, app: express.Application, augur: Au
             websocket.send(makeJsonRpcResponse(message.id, true));
           } else {
             dispatchJsonRpcRequest(db, message as JsonRpcRequest, augur, (err: Error|null, result?: any): void => {
+              if (websocket.readyState !== WebSocket.OPEN ) {
+                console.log("Client disconnected during request, ignoring response");
```
tinybike: This feels more like a `console.warn` to me...thoughts?

tinybike:
Nice find!

src/server/run-websocket-server.ts
epheph: yep! it's not normally a big deal, but it could be useful if you see a lot of them that something is up

PR #223, opened 2018-02-27
Epheph/ch7185/error could not find balance for token decrease

epheph:
@tinybike this fixes that issue you ran into. The Blockstream was setup using an inline data structure that didn't always match the log processor. We were adding this to the log processor, which is used directly during the bulk load, but the blockstream was missing some of those functions. Instead of adding what was missing, I just changed it to re-use the same data structure.

The change itself wasn't large, but log processors had all sorts of mis-named stuff in there we never noticed (but augur.js did, since it checks for matching function)

Here's the meat of this change:
https://github.com/AugurProject/augur-node/compare/epheph/ch7185/error-could-not-find-balance-for-token-decrease?expand=1#diff-ce792255d860090ecfe58ce9d2b95cd4L7


src/blockchain/log-processors/index.ts
```diff
@@ -36,7 +48,7 @@ export const logProcessors: LogProcessors = {
       add: processInitialReporterRedeemedLog,
       remove: processInitialReporterRedeemedLogRemoval,
     },
-    InitialReporterTransferred: {
+    InitialReporterTransfered: {
```
nuevoalex: I think this should be `InitialReporterTransferred` shouldn't it? The typo of one `r` was fixed in augur core.

epheph: Yea i noticed that... hmmm, i must have an old version checked out? I had to make this change to make it work, did this one change after the other ones?

epheph: There hasn't been an augur-core release since -18, which was before you changed that name.

PR #225, opened 2018-02-27
get dispute info by account

epheph:


src/server/getters/get-dispute-info.ts
```diff
@@ -15,6 +16,8 @@ interface DisputesResult {
   markets: Array<MarketsRowWithCreationTime>;
   completedStakes: Array<StakeRow>;
   activeCrowdsourcer: Array<ActiveCrowdsourcer>;
+  accountStakeComplete: Array<any>;
+  accountStakeIncomplete: Array<any>;
   payouts: Array<PayoutRow>;
```
bthaile: can <any> be <string> or nullable string or is it literally any value will work?

epheph: You're referencing the `Array<any>` ? I didn't mean to leave those in there, I'll fix.

PR #228, opened 2018-03-02
advance awaiting fee window

epheph:


src/blockchain/process-block.ts
```diff
@@ -174,12 +174,32 @@ function advanceFeeWindowActive(db: Knex, blockNumber: number, timestamp: number
...
+function advanceAwaitingNextFeeWindow(db: Knex, blockNumber: number, timestamp: number, callback: AsyncCallback) {
+  // Finds crowdsourcers rows that we don't know the completion of, but are attached to feeWindows that have ended
+  // They did not reach their goal, so set completed to 0.
+  getMarketsWithReportingState(db, ["marketID"])
+    .where("reportingState", ReportingState.AWAITING_NEXT_WINDOW)
+    .asCallback((err: Error|null, marketIds: Array<MarketsContractAddressRow>) => {
+      if (err) return callback(err);
+      each(marketIds, (marketIdRow, nextMarketIdRow: ErrorCallback) => {
+        updateMarketState(db, marketIdRow.marketId, blockNumber, ReportingState.CROWDSOURCING_DISPUTE, nextMarketIdRow);
+        augurEmitter.emit("MarketState", {
```
bthaile: wishlist: have augurEmitter use constants that the UI can share same. get away from strings.

epheph: Sounds good, i'll do them all in a  different issue:

https://app.clubhouse.io/augur/stories/workspace/unsaved

PR #233, opened 2018-03-06
Move reporting state during crowdsourcer completion

pgebheim:


epheph:
Perfect, the abstracted market state removal is 🐐

PR #234, opened 2018-03-06
store winning payout, retrieve in markets-info

epheph:
types cleaned up now, this should do what you need  @JohnDanz


epheph:
@JohnDanz OK, the typings here took a lot more work than i expected, but they are in much better shape now. this is ready to merged

JohnDanz:
I agree, feel free to merge, i tested on my end it should work with my pending updates. Once this is merged i'll double check everything then post a PR for the UI updates

PR #235, opened 2018-03-06
on eth disconnect, do not exit, throw an error

epheph:


epheph:
@tinybike this will generate a stack trace on exit (and it's more correct, anyway)

PR #246, opened 2018-03-13
functional change to rename and change 1 var to not be a sum

epheph:


src/types.ts
```diff
@@ -252,10 +252,9 @@ export interface UIDisputeTokens {
 }
 
 export interface StakeDetails extends NormalizedPayout {
-  totalStake: string;
-  completedStake: string;
+  stakeCompleted: string;
   size?: string;
```
bthaile: shouldn't there always be a size? here is nullable.

src/types.ts
```diff
@@ -252,10 +252,9 @@ export interface UIDisputeTokens {
...
-  currentStake?: string;
+  stakeCurrent?: string;
   tentativeWinning: boolean;
```
bthaile: also stakeCurrent, I don't see a reason for it to be nullable, it would just b '0'

src/server/getters/get-dispute-info.ts
```diff
@@ -91,46 +91,46 @@ function reshapeStakeRowToUIStakeInfo(stakeRows: DisputesResult): UIStakeInfo|nu
...
+    const accountStakeCurrent = accountStakeCurrentByPayout[payout.payoutId];
 
-    const completedStakeAmount = new BigNumber(completedStakes == null ? 0 : completedStakes.amountStaked.toString());
-    const totalStakeOnPayout = completedStakeAmount.add(new BigNumber(activeCrowdsourcer == null ? 0 : activeCrowdsourcer.amountStaked.toString()));
+    const stakeCompletedAmount = new BigNumber(stakeCompleted == null ? 0 : stakeCompleted.amountStaked.toString());
+    const stakeCurrentOnPayout = new BigNumber(stakeCurrent == null ? 0 : stakeCurrent.amountStaked.toString());
 
     let currentAmounts: StakeSizes;
     if (payout.tentativeWinning === 1 || !isActiveMarketState(marketRow.reportingState)) {
       currentAmounts = {};
```
bthaile: looks good, no dispute if market isn't in correct state.

bthaile:
Is the key in the works or decide it's not needed?

src/types.ts
epheph: No size if it is the current tentativewinning

src/types.ts
epheph: Again, tentativewinning

epheph:
@bthaile actually,i was wrong! it is already keyed under "stakes", it looks like:
```
[
 {
   marketId,
   stakes: []
  }
]
```
so i can just add more keys at that top level

PR #247, opened 2018-03-14
dispute remaining & total stake

epheph:


src/server/getters/get-dispute-info.ts
```diff
@@ -138,6 +146,7 @@ function reshapeStakeRowToUIStakeInfo(stakeRows: DisputesResult): UIStakeInfo|nu
   const disputeRound = totalCompletedStakeOnAllPayouts.equals(0) ? null : stakeRows.disputeRound[0] == null ? 0 : stakeRows.disputeRound[0].disputeRound;
   return {
     marketId: marketRow.marketId,
+    stakeCompletedTotal: totalCompletedStakeOnAllPayouts.toFixed(),
```
bthaile: This is the total stake for the last completed dispute bond? Or is it the stake needed to complete the current active disptue?

epheph: It is the sum of all currently completed stake, including initial, so you will need to double it. I could double add a double'd version if you like.

bthaile: Could you add a stakeCurrentTotal that would be the current dispute bond total? Does that make sense? 
stakeCurrentTotal would be double the stakeCompletedTotal. So if the market is in awaiting fee window after designated reporter then stakeCompletedTotal would be the DR stake and stakeCurrentTotal would be double stakeCompletedTotal, correct?

bthaile:
I'll run with your current changes. Also is the BigNumber string issue been merged up? I see some of the toString() have been removed.

epheph:
accidental removal of toString, i'll re-add

epheph:
I think we can do without the ones i calculatebondsize, since those have to be bignumber's already

PR #249, opened 2018-03-15
scalar denomination

epheph:


epheph:
@stephensprinkle changing the "extraInfo" response (which is really longDescription) to `details` as you requested

PR #251, opened 2018-03-19
modify universe creation and provide  universe migration totals getter

nuevoalex:
handle the new universe logging data and provide a getter to retrieve migration totals

### DIFF
--- a/src/server/getters/get-fork-migration-totals.ts
+++ b/src/server/getters/get-fork-migration-totals.ts
@@ -1,0 +1,37 @@
+import * as Knex from "knex";
+import Augur from "augur.js";
+import { Address } from "../../types";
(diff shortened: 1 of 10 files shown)


src/server/getters/get-fork-migration-totals.ts
```diff
@@ -0,0 +1,37 @@
+import * as Knex from "knex";
+import Augur from "augur.js";
+import { Address } from "../../types";
+import { ForkMigrationTotalsRow, UIForkMigrationTotals } from "../../types";
+import { normalizePayouts } from "./database";
```
adrake33: Does this actually get used anywhere? I think this can probably be removed.

nuevoalex: Yup! will do

PR #253, opened 2018-03-22
Added raw log fields to all manual emits; removed duplicate emits; ch…

tinybike:
…anged ignoreLog to passThroughLog so all logs are re-emitted

Resolves [ch8164](https://app.clubhouse.io/augur/story/8164/some-events-are-not-getting-re-emitted-some-are-getting-emitted-twice)

### DIFF
--- a/src/blockchain/log-processors/completesets.ts
+++ b/src/blockchain/log-processors/completesets.ts
@@ -19,7 +19,7 @@ export function processCompleteSetsPurchasedOrSoldLog(db: Knex, augur: Augur, lo
       tradeGroupId: log.tradeGroupId,
       numPurchased: log.numCompleteSets,
     };
-    augurEmitter.emit(log.eventName, completeSetPurchasedData);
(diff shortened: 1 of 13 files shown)


epheph:
do you think smashing the log and the structured data into the same object could lead to confusion? Could we have an object with "market" and "marketId"? Might be cleaner to make the `log` a key of the emitted object, but it's probably not too big a deal

tinybike:
> do you think smashing the log and the structured data into the same object could lead to confusion? Could we have an object with "market" and "marketId"? Might be cleaner to make the log a key of the emitted object, but it's probably not too big a deal
The problem with making `log` a key of the emitted object is that it wouldn't be consistent between the auto-emits and the manual emits, unless we made it so _all_ emitted logs were `{ log }`...

epheph:
Or we made our sanitized/structured data the key, that is sometimes there if we did processing. then it's consistent

PR #259, opened 2018-03-27
process market migrated events

nuevoalex:
process the market migrated event


src/blockchain/log-processors/market-migrated.ts
```diff
@@ -0,0 +1,12 @@
+import Augur from "augur.js";
+import * as Knex from "knex";
+import { FormattedEventLog, ErrorCallback } from "../../types";
+import { augurEmitter } from "../../events";
```
adrake33: This line can be removed.

PR #260, opened 2018-03-29
Universes getter

nuevoalex:
[Clubhouse Story](https://app.clubhouse.io/augur/story/8393)


src/server/getters/get-universes-info.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (universe == null || account == null) return callback(new Error("Must provide both universe and account"));
+
+  const query = db.select("parentUniverse").from("universes").where("universe", universe).first();
+
+  query.asCallback((err: Error|null, parentUniverseRow: ParentUniverseRow) => {
+    if (err) return callback(err);
+
+    const query = db.select(["universes.universe", "universes.parentUniverse", "balances.balance", "token_supply.supply", db.raw("count(markets.marketId) as numMarkets")]).from("universes")
+      .where("universes.parentUniverse", universe) // Children
+      .orWhere("universes.parentUniverse", parentUniverseRow.parentUniverse) // Siblings
```
epheph: If you pass in a universe that does not exit, AN crashes, since this dereferences `undefined`

epheph:
Do we emit a UniverseCreated for genesis?

src/server/getters/get-universes-info.ts
```diff
@@ -0,0 +1,39 @@
...
+      .orWhere("universes.universe", universe) // Universe
+      .orWhere("universes.universe", parentUniverseRow.parentUniverse) // Parent
+      .leftJoin("token_supply", "token_supply.token", "universes.reputationToken")
+      .leftJoin("balances", function () {
+        this
+          .on("balances.owner", db.raw('"' + account + '"'))
```
epheph: Should we be concerned about sql injection here?

epheph:
A couple comments, but overall looks great, feel free to merge after those issues are fixed

nuevoalex:
Great point about emitting a log for the genesis universe. Currently we do not. This is an easy change though in the contracts so I'll do that separately.

src/server/getters/get-universes-info.ts
nuevoalex: Will add a test and fix!

src/server/getters/get-universes-info.ts
nuevoalex: Probably. I'll look at the recommended way to do this and avoid sql injection

src/server/getters/get-universes-info.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const query = db.select(["universes.universe", "universes.parentUniverse", "balances.balance", "token_supply.supply", db.raw("count(markets.marketId) as numMarkets")]).from("universes")
+      .where("universes.parentUniverse", universe) // Children
+      .orWhere("universes.parentUniverse", parentUniverseRow.parentUniverse) // Siblings
+      .orWhere("universes.universe", universe) // Universe
+      .orWhere("universes.universe", parentUniverseRow.parentUniverse) // Parent
+      .leftJoin("token_supply", "token_supply.token", "universes.reputationToken")
+      .leftJoin("balances", function () {
+        this
+          .on("balances.owner", db.raw("?", [account]))
```
epheph: Didn't know you could do that, I actually need to do this on my branch, too!

PR #263, opened 2018-03-30
endTime all over

epheph:


epheph:
This will require a UI change, getMarketsInfo now returns endTime, not endDate

JohnDanz:
https://github.com/AugurProject/augur/pull/1360 UI side changes that will need to be merged with this PR

PR #271, opened 2018-04-07
add payout so we know how to describe universes in the UI

nuevoalex:


src/server/getters/get-universes-info.ts
```diff
@@ -28,12 +42,24 @@ export function getUniversesInfo(db: Knex, augur: Augur, universe: Address, acco
...
         })
       .leftJoin("markets", "markets.universe", "universes.universe")
+      .leftJoin("payouts", "payouts.payoutId", "universes.payoutId")
       .groupBy("universes.universe");
 
     query.asCallback((err: Error|null, universeInfoRows: Array<UniverseInfoRow<BigNumber>>): void => {
       if (err) return callback(err);
       callback(null, universeInfoRows.map((row: UniverseInfoRow<BigNumber>) => {
-        return formatBigNumberAsFixed(row);
+        const payout: Array<string> = [
```
epheph: We have a function for this, `normalizePayouts`, but we can clean it up later, looks like it behaves the same.

PR #276, opened 2018-04-13
Refactor current fee window

epheph:
The gist of this change is that we are no longer tracking a Feewindow's start and end blocks, just whether it isActive or not. Then the processBlock checks to see if we need to active a new Feewindow or not. 

With this change, we also move markets to the next feeWindow NOT when the FeeWindow is active, but when the event that completes its current phase is done (Crowdsource or Initial Report)


bthaile:
We need to merge this so we can test it.

PR #280, opened 2018-04-18
Petong/travis ci

pgebheim:
Auto build stable docker on push to stable branch


scripts/release-stable.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -x
+
+version=$(date -u +%Y-%m-%d-%H%M)
+docker build . --tag augurproject/augur-node:stable --tag augurproject/augur-node:$version || exit 1
```
epheph: set -e if we're going to `exit 1` on each? no biggie tho

PR #282, opened 2018-04-19
Generalize fetching positions in market

epheph:


epheph:
Just sets it up to be used independent from db updates, doesn't change any functionality

Issue #293, opened 2018-04-29
One Click Digital Ocean Deploy


tomkysar:
Augur Node should be deployable with Digital Ocean's one-click deploy.

Please join the #dev or #bounties channel of our [Discord](http://augur.net/invite) with any questions. 

Bounty posted with [Gitcoin](https://gitcoin.co).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 DAI (300.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/AugurProject/augur-node/293/332).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $6,678.38 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started on the 300.0 DAI (300.0 USD @ $1.0/DAI) funding by__: 
 1. [@kylewu](https://gitcoin.co/profile/kylewu)  

 On the above list? Please leave a comment to let the funder (@tomkysar) know your  plan. If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/AugurProject/augur-node/293/332)
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $7,279.94 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 300.0 DAI (300.0 USD @ $1.0/DAI) has been submitted by__: 
 1. 
 1. [@kylewu](https://gitcoin.co/profile/kylewu) 

 

@tomkysar please take a look at the submitted work:
* [PR](https://github.com/AugurProject/augur-node/pull/295) by @kylewu
 

[...]

PR #295, opened 2018-04-30
DigitalOcean one click deploy

kylewu:
https://github.com/AugurProject/augur-node/issues/293

Not like heroku, i could not find one button deployment support from DO.
Only 3rd party service, like https://installer.71m.us, supports deploy in DO with app.yml configuration.
Please check `https://github.com/kylewu/augur-node` and there is one `Install on DigitalOcean` button on top of README.

NOTE
===
installer.71m.us is HTTPS, when it checks our droplet status, it sends request through http to ip of the newly created droplet. This request will be blocked because of security reason (see browser console for detail). 
one solution is to deploy `installer.71m.us` yourself (check its page for detail)

Verification
===
curl  46.101.77.91:9001
Hello World

`46.101.77.91` is my droplet

### DIFF
--- a/app.yml
+++ b/app.yml
@@ -1,0 +1,16 @@
+name: augur-node
+image: ubuntu-16-04-x64
+min_size: 1gb
+config:
+  #cloud-config
+  runcmd:
+    - apt-get install -y build-essential
+    - apt-get install -y git
+    - apt-get install -y sqlite3
+    - echo "export ENDPOINT_HTTP=https://rinkeby.ethereum.nodes.augur.net" | tee -a /root/.bashrc
+    - echo "export ENDPOINT_WS=wss://websocket-rinkeby.ethereum.nodes.augur.net" | tee -a /root/.bashrc
+    - curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
+    - apt-get install -y nodejs
+    - cd /root && git clone https://github.com/kylewu/augur-node.git 
+    - cd /root/augur-node && npm install
+    - cd /root/augur-node && export ENDPOINT_HTTP=https://rinkeby.ethereum.nodes.augur.net && export ENDPOINT_WS=wss://websocket-rinkeby.ethereum.nodes.augur.net && npm run clean-start
--- a/src/blockchain/sync-augur-node-with-blockchain.ts
+++ b/src/blockchain/sync-augur-node-with-blockchain.ts
@@ -55,8 +55,22 @@ function monitorEthereumNodeHealth(augur: Augur) {
   }, 5000);
 }
 
+const getHttpUrl = (network: NetworkConfiguration) => {
+  if (process.env.ENDPOINT_HTTP) {
+    return process.env.ENDPOINT_HTTP;
+  }
+  return network.http;
+};
+
+const getWsUrl = (network: NetworkConfiguration) => {
+  if (process.env.ENDPOINT_WS) {
+    return process.env.ENDPOINT_WS;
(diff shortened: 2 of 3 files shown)


src/blockchain/sync-augur-node-with-blockchain.ts
```diff
@@ -55,8 +55,22 @@ function monitorEthereumNodeHealth(augur: Augur) {
   }, 5000);
 }
 
+const getHttpUrl = (network: NetworkConfiguration) => {
```
pgebheim: This isn't necessary.

NetworkConfiguration will read from the environment by default, so if you run without a NetworkConfiguration, `network.ws === process.env.ETHEREUM_HTTP`.

Just use ETHEREUM_HTTP and ETHEREUM_WS in place of the old ENDPOINT_HTTP and ENDPOINT_WS.

src/blockchain/sync-augur-node-with-blockchain.ts
```diff
@@ -55,8 +55,22 @@ function monitorEthereumNodeHealth(augur: Augur) {
...
+const getWsUrl = (network: NetworkConfiguration) => {
+  if (process.env.ENDPOINT_WS) {
+    return process.env.ENDPOINT_WS;
+  }
+  return network.ws;
+};
+
 export function syncAugurNodeWithBlockchain(db: Knex, augur: Augur, network: NetworkConfiguration, uploadBlockNumbers: UploadBlockNumbers, callback: ErrorCallback): void {
-  augur.connect({ ethereumNode: { http: network.http, ws: network.ws }, startBlockStreamOnConnect: false }, (): void => {
+  augur.connect({ ethereumNode: { http: getHttpUrl(network), ws: getWsUrl(network) }, startBlockStreamOnConnect: false }, (): void => {
```
pgebheim: Falls into the same category as above.

app.yml
```diff
@@ -0,0 +1,16 @@
+name: augur-node
```
pgebheim: I'd encapsulate all this functionality in a script and put it in /scripts, see docket-entrypoint.sh.

Even better would be to have DO just run the docker image.

pgebheim:
Closing without merging -- I am definitely open to a script that can run on DO if provided credentials or a more supported service for doing the deploys.

kylewu:
@pgebheim thanks for your comments
do you prefer shell script, python script or node? I'm planning to create another pr

sharkcrayon:
@kylewu Sorry for the delay in addressing this. We'll be getting back to you next week on the best way to move forward.

petong:
@kylewu The original idea was that the bounty for this task be a pre-built application on Digital Ocean and not a shell/python/node script. See examples here: https://www.digitalocean.com/products/one-click-apps/

kylewu:
@petong yes, this PR is not the script.

@everyone please kindly follow the conversation and try to understand what happend here. This is the first PR I created and it’s about the one button deployment while the other (later) is about the script.

I did this in the beginning, then I was told a script is preferable. After the script PR was created I was told one button deployment is required.

If you believe one button is what you need, please leave your review in this PR. It won’t take more than 5 minutes to check code.

kylewu:
I’m leaving this conversation because I don’t have enough time to follow this. I hope someone could continue. Thanks for your time.

pgebheim:
Closing this PR, but I've taken this script and adapted it to be a bit more robust:

https://github.com/AugurProject/augur-node/pull/589

Thanks for the contrib @kylewu

PR #299, opened 2018-05-02
Add markets.forking and markets.needsMigration fields

epheph:


src/migrations/20171005123235_markets.ts
```diff
@@ -25,6 +25,8 @@ exports.up = async (knex: Knex): Promise<any> => {
       "feeWindow" varchar(66),
       "endTime" integer NOT NULL CONSTRAINT "positiveEndTime" CHECK ("endTime" > 0),
       "finalizationTime" integer,
+      "forking" boolean DEFAULT O,
```
adrake33: I think this may need to be changed. It looks like this is defaulting to the letter "O" instead of zero?

epheph: 🤦‍♂️

adrake33:
Took a look at the latest changes. Looks good!

PR #308, opened 2018-05-04
Add outcome.sharesToken, stuff off-chain call into DB

epheph:
we'll need this to look up balances of winning shares


epheph:
This isn't the getter yet, but it's part of blockchain processing required to get it working.

epheph:
Just noticed this is already available in `tokens` table if you filter by `shares`

PR #309, opened 2018-05-05
Modified output of `getReportingFees`to return forked market & nonforked markets

adrake33:
To see the new format of the output, see https://app.clubhouse.io/augur/story/9292/add-more-fields-to-object-returned-by-augur-reporting-getreportingfees.


src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +101,127 @@ interface RepStakeResults {
...
+  const query = db("fee_windows")
+  .select("universes.universe", "universes.parentUniverse")
+  .join("universes", "fee_windows.universe", "universes.universe")
+  .groupBy("fee_windows.universe");
+  if (universe != null) {
+    query.where("fee_windows.universe", universe);
+  } else if (feeWindow != null) {
+    query.where("fee_windows.feeWindow", feeWindow);
+  }
+  query.asCallback((err, universeAndParentUniverse: Array<UniverseAndParentUniverse>) => {
```
epheph: instead of `select()` use `first()`. It does a `LIMIT 1` and makes your callback value a single object (or null) instead of an array

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +101,127 @@ interface RepStakeResults {
   lostRep: BigNumber;
 }
 
+function getUniverseAndParentUniverse(db: Knex, reporter: Address, universe: Address|null, feeWindow: Address|null, callback: (err: Error|null, result?: UniverseAndParentUniverse) => void) {
```
epheph: reporter argument unused

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +101,127 @@ interface RepStakeResults {
...
+  for (const key in keyedNonforkedMarkets) {
+    if (keyedNonforkedMarkets.hasOwnProperty(key)) {
+      nonforkedMarkets.push(keyedNonforkedMarkets[key]);
+    }
+  }
+
+  return callback(null, {forkedMarket, nonforkedMarkets});
+}
+
+function getMarkets(db: Knex, reporter: Address, universe: Address, parentUniverse: Address, callback: (err: Error|null, formattedMarketInfo?: FormattedMarketInfo) => void) {
```
epheph: getMarkets is the same name as the getter used for getMarkets. There's no namespace problems, but it might be nice to use a different name so there's no confusion

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +101,127 @@ interface RepStakeResults {
...
+  const initialReportersQuery = db("initial_reports")
+    .select(["initial_reports.initialReporter", "initial_reports.disavowed", "initial_reports.marketId", "markets.universe", "market_state.reportingState", "markets.forking", "markets.needsMigration"])
+    .join("markets", "initial_reports.marketId", "markets.marketId")
+    .join("market_state", "market_state.marketStateId", "markets.marketStateId")
+    .whereIn("markets.universe", [universe, parentUniverse])
+    .andWhere((queryBuilder) => queryBuilder.where("market_state.reportingState", "AWAITING_FINALIZATION").orWhere("market_state.reportingState", "FINALIZED"))
```
epheph: `whereIn()` might be a nicer way to do this?

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +101,127 @@ interface RepStakeResults {
...
+    query.where("fee_windows.feeWindow", feeWindow);
+  }
+  query.asCallback((err, universeAndParentUniverse: Array<UniverseAndParentUniverse>) => {
+    if (err) return callback(err);
+    return callback(null, universeAndParentUniverse[0]);
+  });
+}
+
+function formatMarketInfo(marketParticipants: any, parentUniverse: Address, callback: (err: Error|null, formattedMarketInfo?: FormattedMarketInfo) => void) {
+  let forkedMarket = {} as ForkedMarket;
```
epheph: This isn't an async call, can we convert it to a non-callback call?

adrake33: Yeah, this shouldn't be an async call. I'll change it.

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +98,124 @@ interface RepStakeResults {
...
+    if (keyedNonforkedMarkets.hasOwnProperty(key)) {
+      nonforkedMarkets.push(keyedNonforkedMarkets[key]);
+    }
+  }
+
+  return {forkedMarket, nonforkedMarkets};
+}
+
+function getMarketsReportingParticipants(db: Knex, reporter: Address, universe: Address, parentUniverse: Address, callback: (err: Error|null, formattedMarketInfo?: FormattedMarketInfo) => void) {
+  const initialReportersQuery = db("initial_reports")
```
epheph: Should initialReporters be filtered down to who the reporter is?

adrake33: Yep, just added this.

src/server/getters/get-reporting-fees.ts
```diff
@@ -4,22 +4,58 @@ import * as _ from "lodash";
...
     unclaimedRepEarned: string;
     lostRep: string;
-    claimedEth: string;
-    claimedRepStaked: string;
-    claimedRepEarned: string;
   };
   feeWindows: Array<Address>;
-  crowdsourcers: Array<Address>;
-  initialReporters: Array<Address>;
+  forkedMarket: ForkedMarket;
```
epheph: This is not always present right? The type implies it is. In the code, you specify `{} as ForkedMarket` which coerces TypeScript a bit, and allows this incorrect definition (which implies there is always a ForkedMarket here, and a ForkedMarket always has required structure). Something like this would help:


```
  let forkedMarket: ForkedMarket|null = null;
```

Then the type could be :
```
forkedMarket: ForkedMarket|null;
```

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +98,124 @@ interface RepStakeResults {
...
+          isForked: marketParticipants.initialReporters[i].disavowed ? true : false,
+        },
+      };
+    } else if (!marketParticipants.initialReporters[i].forking && marketParticipants.initialReporters[i].universe !== parentUniverse) {
+      keyedNonforkedMarkets[marketParticipants.initialReporters[i].marketId] = {
+        address: marketParticipants.initialReporters[i].marketId,
+        universeAddress: marketParticipants.initialReporters[i].universe,
+        crowdsourcersAreDisavowed: false,
+        isMigrated: !marketParticipants.initialReporters[i].needsMigration,
+        isFinalized: marketParticipants.initialReporters[i].reportingState === "FINALIZED",
```
epheph: There is an enum for these, `ReportingState` , can you replace all occurrences?

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +98,128 @@ interface RepStakeResults {
...
+        isFinalized: marketParticipants.initialReporters[i].reportingState === augur.constants.REPORTING_STATE.FINALIZED,
+        crowdsourcers: [],
+        initialReporterAddress: marketParticipants.initialReporters[i].initialReporter,
+      };
+    }
+  }
+  for (i = 0; i < marketParticipants.crowdsourcers.length; i++) {
+    if (marketParticipants.crowdsourcers[i].forking && marketParticipants.crowdsourcers[i].universe === parentUniverse) {
+      let forkedMarketIsEmpty: boolean = true;
+      if (forkedMarket) {
```
epheph: Now that forked market is either null OR an object, we can just check it for `null` without iterating its keys

src/server/getters/get-reporting-fees.ts
```diff
@@ -62,6 +98,128 @@ interface RepStakeResults {
...
+          crowdsourcers: [marketParticipants.crowdsourcers[i].crowdsourcerId],
+          initialReporterAddress: null,
+        };
+      } else {
+        keyedNonforkedMarkets[marketParticipants.crowdsourcers[i].marketId].crowdsourcersAreDisavowed = marketParticipants.crowdsourcers[i].disavowed ? true : false;
+        keyedNonforkedMarkets[marketParticipants.crowdsourcers[i].marketId].crowdsourcers.push(marketParticipants.crowdsourcers[i].crowdsourcerId);
+      }
+    }
+  }
+  const nonforkedMarkets: Array<NonforkedMarket> = [];
```
epheph: Could this be replaced with 
`const nonforkedMarkets = Object.values(keyedNonforkedMarkets)` ?

src/server/getters/get-reporting-fees.ts
adrake33: Oops, good point. I forgot to take out the part where it's iterating over the keys. Will remove that.

src/server/getters/get-reporting-fees.ts
adrake33: Hmm, I tried that, but the linter was saying "Property `values` does not exists on type `ObjectConstructor`".

I was able to accomplish this using the following instead, though:
```
const nonforkedMarkets = Object.keys(keyedNonforkedMarkets).map(function(key) { return keyedNonforkedMarkets[key]; });
```

epheph: You're right, Object.values is es2017. That line you put above looks like the next best thing.

PR #310, opened 2018-05-05
Monitor health on reset

nuevoalex:
This was the only piece making hot network switching not work in augur-app as it turns out. Really just made it throw errors on a cadence but everything else worked so it was perplexing


nuevoalex:
The DB is being destroyed and rebuilt.
The other concerns seem to be working correctly by just starting up a new connection / instance.

PR #312, opened 2018-05-06
lower case addresses we are already pre-processing

epheph:


src/server/address-format-reviver.ts
```diff
@@ -22,9 +22,12 @@ export const inputsExpectedAsAddress: StringToBoolMap = {
 };
 
 export function addressFormatReviver (key: string, value: any) {
-  const valueIsOfAddressType: boolean = typeof value === "string" || Array.isArray(value);
-  if (valueIsOfAddressType && inputsExpectedAsAddress[key] === true) {
-    return formatEthereumAddress(value);
+  if (inputsExpectedAsAddress[key] === true) {
+    if ( typeof value === "string" ) {
+      return formatEthereumAddress(value.toLowerCase());
```
bthaile: Do you also need a trim? Just in case whitespace

epheph: Sure, doesn't hurt, added

PR #320, opened 2018-05-09
Handle errors

nuevoalex:
This should address https://app.clubhouse.io/augur/story/9296/augur-node-epipe-error


epheph:
Think this could be cause of EPIPE?

nuevoalex:
I think so! Apparently if you don't explicitly handle these is causes stack traces similar to the one in the issue description

PR #321, opened 2018-05-09
getwinningbalance getter

epheph:


epheph:
Yes, attoeth. I'm not taking fees out yet, though, I probably should? I can make in next PR

PR #332, opened 2018-05-11
take universe as part of the initial reporters request

nuevoalex:
It was trying to get a REP contract address from the ether and was never working probably. Now it takes a universe and gets the rep contract associated with it.

### DIFF
--- a/src/server/getters/get-initial-reporters.ts
+++ b/src/server/getters/get-initial-reporters.ts
@@ -4,14 +4,13 @@ import { Address } from "../../types";
 import { formatBigNumberAsFixed } from "../../utils/format-big-number-as-fixed";
 import { InitialReportersRow, UIInitialReporters } from "../../types";
 
-export function getInitialReporters(db: Knex, augur: Augur, reporter: Address, redeemed: boolean|null|undefined, withRepBalance: boolean|null|undefined, callback: (err: Error|null, result?: any) => void): void {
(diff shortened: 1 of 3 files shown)


src/server/getters/get-initial-reporters.ts
```diff
@@ -4,14 +4,13 @@ import { Address } from "../../types";
...
-export function getInitialReporters(db: Knex, augur: Augur, reporter: Address, redeemed: boolean|null|undefined, withRepBalance: boolean|null|undefined, callback: (err: Error|null, result?: any) => void): void {
+export function getInitialReporters(db: Knex, augur: Augur, universe: Address, reporter: Address, redeemed: boolean|null|undefined, withRepBalance: boolean|null|undefined, callback: (err: Error|null, result?: any) => void): void {
   let query = db("initial_reports")
     .select(["marketID", "reporter", "amountStaked", "initialReporter", "redeemed", "isDesignatedReporter", "balances.balance AS repBalance"])
     .join("balances", "balances.owner", "=", "initial_reports.initialReporter")
-    .where({
-      reporter,
-      "balances.token": augur.contracts.addresses[augur.rpc.getNetworkID()].ReputationToken,
-    });
+    .join("universes", "universes.universe", db.raw("?", universe))
```
epheph: is there a reason to join this way? It seems backwards to join on a static value, but then filter on a join. If these were swapped, neither would need a db.raw()

nuevoalex: Great point! Will make that change

PR #334, opened 2018-05-12
add "needsDisavowal" for markets and make "disavowed" account for cases

nuevoalex:
As discussed in Discord I didn't make `disavowed` an enum and instead kept it as boolean since the one case we needed that implicitly has the state of "needs disavowal" (forked market + not disavowed).

I added `needsDisavowal` for markets in a forking universe that haven't been migrated or disavowed.

Migration and manual disavowal will now mark all the market's crowdsourcers as disavowed as well.

### DIFF
--- a/src/blockchain/log-processors/market-migrated.ts
+++ b/src/blockchain/log-processors/market-migrated.ts
@@ -3,9 +3,27 @@ import * as Knex from "knex";
 import { FormattedEventLog, ErrorCallback } from "../../types";
 
 export function processMarketMigratedLog(db: Knex, augur: Augur, log: FormattedEventLog, callback: ErrorCallback): void {
-  db.update({ universe: log.newUniverse, needsMigration: db.raw("needsMigration - 1") }).into("markets").where("marketId", log.market).asCallback(callback);
+  db.update({
+    universe: log.newUniverse,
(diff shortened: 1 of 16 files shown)


src/blockchain/log-processors/market-migrated.ts
```diff
@@ -3,9 +3,27 @@ import * as Knex from "knex";
 import { FormattedEventLog, ErrorCallback } from "../../types";
 
 export function processMarketMigratedLog(db: Knex, augur: Augur, log: FormattedEventLog, callback: ErrorCallback): void {
-  db.update({ universe: log.newUniverse, needsMigration: db.raw("needsMigration - 1") }).into("markets").where("marketId", log.market).asCallback(callback);
+  db.update({
+    universe: log.newUniverse,
+    needsMigration: db.raw("needsMigration - 1"),
+    needsDisavowal: db.raw("needsDisavowal - 1"),
```
epheph: Is this necessary for needsDisavowal? I did it for migration, because a market might need to be migrated through more than one fork, can the same thing happen for disavow?

nuevoalex: I thought this was to support rollback since it is possible the market already needed disavowal from a previous fork and we don't want to set it off on rollback in that case.

PR #336, opened 2018-05-13
issue #293 digital ocean deploy script

kylewu:
#293

### DIFF
--- a/scripts/do-deploy.sh
+++ b/scripts/do-deploy.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+[ -z "$DIGITALOCEAN_ACCESS_TOKEN" ] && echo "Need to set DIGITALOCEAN_ACCESS_TOKEN" && exit 1;
+
+echo "preparing env"
+
+# create droplet augur-node
+docker-machine inspect augur-node > /dev/null 2>&1 || docker-machine create --driver digitalocean augur-node
+
+eval $(docker-machine env augur-node)
+
+# kill augur-node if running
+docker kill augur-node 2>/dev/null
+
+echo "running docker"
+# run docker in droplet
+docker run -d --rm -p 9001:9001 -p 9002:9002 -e "ETHEREUM_WS=wss://websocket-rinkeby.ethereum.nodes.augur.net" --name augur-node augurproject/augur-node:stable
+
+echo "http://$(docker-machine ip augur-node):9001 is up"


kylewu:
@pgebheim please take a look at this PR. It includes a script to run augur-node in digital ocean

MicahZoltu:
I believe the goal with this is to have a deploy to digital ocean button, similar to the Heroku Button and Azure Button seen here:
https://blog.heroku.com/heroku-button
https://azure.microsoft.com/en-us/blog/deploy-to-azure-button-for-azure-websites-2/

I am not sure if there are any tutorials for something similar for Digital Ocean or even if it is possible, but I believe that is what the bounty is for rather than a shell script.

MicahZoltu:
Looks like it is possible, this site does it somehow (note the button at the top): http://apostello.readthedocs.io/en/latest/deploy_do.html

pgebheim:
@petong  -- Can you take a look at this from an ops perspective and see if we wanna include it as a quickstart for a remote hosted Augur Node?

kylewu:
@MicahZoltu  in my first PR https://github.com/AugurProject/augur-node/pull/295 i took the approach as you describe. but i was told a script is better. please check the comment in #295

MicahZoltu:
Hmm, @pgebheim your argument against the original PR goes against the original design goals for what we wanted from these buttons.  Perhaps the design goals have changed, but the original goal assumed that our target audience is non-tech savvy users, which means anything that requires running something from a command prompt is a non-starter, especially if it requires linux (e.g., shell script).  The deploy to Heroku, Deploy to Azure, Deploy to Digital Ocean buttons were meant to do everything the user needed via a web interface, and ideally without requiring the user copy around access keys.  I know this is possible with Heroku and Azure one-click deploys, and the example in the link I provided implies that it is possible with Digital Ocean as well.

I would argue that a deployment script isn't really worth having, as anyone that is tech savvy enough to use that can just host the docker container pretty trivially.

I'm reasonably confident that the original PR was much closer to what is required for the bounty than a shell script.  The "one click deploy" is a term of art that is used throughout the industry to mean those Deploy to Heroku, Deploy to Azure, Deploy to EC2 buttons you see all over the place, not shell scripts (commonly referred to as "turn-key solution").

kylewu:
@MicahZoltu thanks for your comment.

I do not have permission to reopen a pull request, could next reviewer reopen #295 and leave comment there? Thanks!

MicahZoltu:
@kylewu I think we need to make sure that all parties involved properly understand what the goal of this issue is and what the requirements for the bounty are.  At the least, there is a disconnect between Paul and I, and I would hate for you to waste a bunch of time jumping back and forth between two solutions.

Ultimately, @tomkysar and/or @tinybike are the ultimate authority, with @pgebheim being next in line I believe.  I piped up here only because when I discussed this in the past with @tinybike, a shell script wasn't what was desired and I want to get to the bottom of what changed.

kylewu:
@MicahZoltu makes sense. i'll follow this issue and wait for conclusion before i continue.
Anyway, i'm very happy to see some comments here :)

tinybike:
@MicahZoltu is correct -- this issue is specifically to create a one-click deploy (a button); unfortunately, @pgebheim's comment on https://github.com/AugurProject/augur-node/pull/295 is incorrect.

kylewu:
@tinybike thanks for the comment. I did the one-click deploy here https://github.com/AugurProject/augur-node/pull/295 in the beginning

tinybike:
So this is or isn't a one click deploy?  I understood this to be a deploy script.  It needs to be a button.

kylewu:
@tinybike please check PR https://github.com/AugurProject/augur-node/pull/295
it's closed by pg. I have no permission to reopen it

kylewu:
I'm going to close this pr, please go to https://github.com/AugurProject/augur-node/pull/295

PR #338, opened 2018-05-13
always return forkedMarket with getReportingFees

nuevoalex:


nuevoalex:
Also includes the isForked -> needsFork change

PR #339, opened 2018-05-13
Update tests to match augur.js and fix calculations

pgebheim:


src/server/getters/get-profit-loss.ts
```diff
@@ -23,16 +23,17 @@ export interface PLBucket {
   profitLoss?: ProfitLoss | null;
 }
 
-export function calculateBucketProfitLoss(augur: Augur, trades: Array<TradingHistoryRow>, buckets: Array<PLBucket>): Array<PLBucket> {
+export type TradeRow = TradingHistoryRow & { type: string, maker: boolean};
```
tinybike: I'm not familiar with this syntax; what does the single `&` mean here?

src/server/getters/get-profit-loss.ts
```diff
@@ -126,9 +127,16 @@ function sumProfitLossResults(left: PLBucket, right: PLBucket): PLBucket {
...
+  const tradeHistory: Array<TradingHistoryRow> = await queryTradingHistory(db, universe, account, null, null, null, null, endTime, "trades.blockNumber", false)
     .orderBy("trades.marketId")
     .orderBy("trades.outcome");
+    
+  const trades: Array<TradeRow> = tradeHistory.map((row: TradingHistoryRow): TradeRow => {
+      return Object.assign({}, row, {
+        type: row.orderType!,
+        maker: account === row.creator!
+      });
+    }) as Array<TradeRow>;
```
tinybike: Is the `as Array<TradeRow>` necessary?

test/server/getters/get-profit-loss.js
```diff
@@ -2,11 +2,20 @@ const Augur = require("augur.js");
...
+const sqlite3 = require("sqlite3");
+const Knex = require("knex");
+const { postProcessDatabaseResults } = require("../../../build/server/post-process-database-results");
 
 const START_TIME = 1506474500;
 const MINUTE_SECONDS = 60;
 const HOUR_SECONDS = MINUTE_SECONDS*60;
 
+const BigNumber = require('bignumber.js');
+BigNumber.config({
```
tinybike: Should we also call `BigNumber.config` in the source files?  I know this ended up not being an issue earlier, but it would be good to make sure `BigNumber` here and in augur.js are configured the same way.

test/server/getters/get-profit-loss.js
```diff
@@ -89,6 +98,81 @@ describe("server/getters/get-profit-loss#bucketRangeByInterval", () => {
...
+          "profitLoss": {
+            "meanOpenPrice": "0",
+            "position": "0",
+            "realized": "4",
+            "total": "4",
+            "unrealized": "0",
+          },
+          "timestamp": endTime,
+        }]);
+      } catch(e) {
```
tinybike: Hm, why use `try { ... } catch (e) { done(e) }` instead of just allowing the `assert.deepEqual` to throw?

test/server/getters/get-profit-loss.js
```diff
@@ -595,15 +670,93 @@ describe("server/getters/get-profit-loss", () => {
...
+
+    const buckets = [{
+      timestamp: 10000,
+      lastPrice: null,
+    }, {
+      timestamp: 20000,
+      lastPrice: 0.1,
+    }];
+    const results = calculateBucketProfitLoss(augur, trades, buckets);
+    const results2 = augur.trading.calculateProfitLoss( { trades, lastPrice: 0.1} )
```
tinybike: Nit: brace spacing doesn't match

src/server/getters/get-profit-loss.ts
pgebheim: It's a union type:

It's TradingHistoryRow plus a type field, and a maker field.

test/server/getters/get-profit-loss.js
pgebheim: You're right we should move this, I'll update.

test/server/getters/get-profit-loss.js
pgebheim: Oh yeah! fixing

test/server/getters/get-profit-loss.js
pgebheim: Because it throws from an async call which doesnt catch and call done

PR #344, opened 2018-05-14
Fix getReportingFees eth calculations

epheph:
These were just not correct and missing all sorts of information, such as cash that was already on the reporting participant, and a reporting participant having spanned multiple windows


src/server/getters/get-reporting-fees.ts
```diff
@@ -157,13 +155,13 @@ function formatMarketInfo(marketParticipants: MarketParticipantRows) {
...
   }
-  const keyedNonforkedMarkets: {[marketId: string]: NonforkedMarket} = {};
+  const keyedNonforkedMarkets: { [marketId: string]: NonforkedMarket } = {};
   let i: number;
   for (i = 0; i < marketParticipants.initialReporters.length; i++) {
     if (marketParticipants.initialReporters[i].forking) {
       forkedMarket.initialReporter = {
         initialReporterId: marketParticipants.initialReporters[i].initialReporter,
-        needsFork: marketParticipants.initialReporters[i].disavowed ? false : true,
+        needsFork: !marketParticipants.initialReporters[i].disavowed,
```
adrake33: It looks like `initial_reports.disavowed` is an integer, so I think this should be "!!"

src/server/getters/get-reporting-fees.ts
```diff
@@ -179,28 +177,28 @@ function formatMarketInfo(marketParticipants: MarketParticipantRows) {
...
     if (marketParticipants.crowdsourcers[i].forking) {
-      forkedMarket.crowdsourcers.push({crowdsourcerId: marketParticipants.crowdsourcers[i].crowdsourcerId, needsFork: marketParticipants.crowdsourcers[i].disavowed ? false : true});
+      forkedMarket.crowdsourcers.push({ crowdsourcerId: marketParticipants.crowdsourcers[i].crowdsourcerId, needsFork: !marketParticipants.crowdsourcers[i].disavowed });
     } else {
       if (!keyedNonforkedMarkets[marketParticipants.crowdsourcers[i].marketId]) {
         keyedNonforkedMarkets[marketParticipants.crowdsourcers[i].marketId] = {
           marketId: marketParticipants.crowdsourcers[i].marketId,
           universe: marketParticipants.crowdsourcers[i].universe,
-          crowdsourcersAreDisavowed: marketParticipants.crowdsourcers[i].disavowed ? true : false,
+          crowdsourcersAreDisavowed: !!marketParticipants.crowdsourcers[i].disavowed,
```
adrake33: And `crowdsourcers.disavowed is a boolean, so I think this line and the one below should use "!"

src/server/getters/get-reporting-fees.ts
```diff
@@ -306,87 +279,125 @@ function getStakedRepResults(db: Knex, reporter: Address, universe: Address|null
...
             const response = {
               total: {
-                unclaimedEth: unclaimedEth.toFixed(),
-                unclaimedRepStaked: repStakeResults.fees.unclaimedRepStaked.toFixed(),
-                unclaimedRepEarned: repStakeResults.fees.unclaimedRepEarned.toFixed(),
-                lostRep: repStakeResults.fees.lostRep.toFixed(),
+                unclaimedEth: unclaimedParticipantEthFees.plus(unclaimedParticipationTokenEthFees).toFixed(0, BigNumber.ROUND_DOWN),
+                unclaimedRepStaked: repStakeResults.fees.unclaimedRepStaked.toFixed(0, BigNumber.ROUND_DOWN),
+                unclaimedRepEarned: repStakeResults.fees.unclaimedRepEarned.toFixed(0, BigNumber.ROUND_DOWN),
+                lostRep: repStakeResults.fees.lostRep.toFixed(0, BigNumber.ROUND_DOWN),
```
adrake33: For the UI, we need to differentiate between the unclaimed ETH/REP related to the forked market, and all other reporting fees (since there are 2 separate buttons for this). Therefore, I think we'll also need `unclaimedForkRepStaked` and `unclaimedForkEthFees` properties to be returned.

PR #347, opened 2018-05-15
don't  send nulls for numbers fields

nuevoalex:


src/seeds/test/balances.ts
```diff
@@ -99,7 +99,7 @@ exports.seed = async (knex: Knex): Promise<any> => {
     }, {
       token: "REP_TOKEN_SECOND_GRAND_CHILD",
       owner: "0x0000000000000000000000000000000000000021",
-      balance: "4",
```
epheph: Would this have emitted null previously if it was 0? Or would it have to not have a row?

nuevoalex: null since there is still a universe balance and other data

PR #354, opened 2018-05-16
getter for escape hatch orders

nuevoalex:
Really simple getter to handle getting orders for use in the escape hatch

### DIFF
--- a/src/server/getters/get-all-orders.ts
+++ b/src/server/getters/get-all-orders.ts
@@ -1,0 +1,18 @@
+import * as Knex from "knex";
+import { Address, AllOrdersRow } from "../../types";
+import { formatBigNumberAsFixed } from "../../utils/format-big-number-as-fixed";
+
+export function getAllOrders(db: Knex, account: Address, callback: (err: Error|null, result?: any) => void): void {
+  const query = db.select(["orderId", "tokensEscrowed", "sharesEscrowed"]).from("orders").where("orderCreator", account);
+  query.asCallback((err: Error|null, allOrders?: Array<AllOrdersRow<BigNumber>>): void => {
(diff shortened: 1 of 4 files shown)


src/server/getters/get-all-orders.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface AllOrders {
+  [orderId: string]: AllOrdersRow<string>;
+}
+
+export function getAllOrders(db: Knex, account: Address, callback: (err: Error|null, result?: any) => void): void {
+  const query = db.select(["orderId", "tokensEscrowed", "sharesEscrowed"]).from("orders").where("orderCreator", account);
+  query.asCallback((err: Error|null, allOrders?: Array<AllOrdersRow<BigNumber>>): void => {
+    if (err) return callback(err);
+    if (!allOrders) return callback(err, []);
```
epheph: Response is an object, we should return {} in the empty case

src/server/getters/get-all-orders.ts
```diff
@@ -0,0 +1,23 @@
+import * as Knex from "knex";
+import { Address, AllOrdersRow } from "../../types";
+import { formatBigNumberAsFixed } from "../../utils/format-big-number-as-fixed";
+
+export interface AllOrders {
+  [orderId: string]: AllOrdersRow<string>;
+}
+
+export function getAllOrders(db: Knex, account: Address, callback: (err: Error|null, result?: any) => void): void {
```
epheph: Throw `AllOrder` type on the result?

src/server/getters/get-all-orders.ts
nuevoalex: good catch

PR #359, opened 2018-05-16
fix for initial reporter unclaimed earned rep

nuevoalex:


src/server/getters/get-reporting-fees.ts
```diff
@@ -265,7 +271,11 @@ function getStakedRepResults(db: Knex, reporter: Address, universe: Address, cal
     initialReporters: (next: AsyncCallback) => initialReportersQuery.asCallback(next),
   }, (err: Error|null, result: StakedRepResults) => {
     if (err) return callback(err);
+
+    const marketDisputed: MarketsDisputedMap = {};
+
     let fees = _.reduce(result.crowdsourcers, (acc: RepStakeResults, feeWindowCompletionStake: FeeWindowCompletionStakeRow) => {
+      marketDisputed[feeWindowCompletionStake.marketId] = true;
```
epheph: Aren't we only returning crowdsourcers for which the user has a stake?

nuevoalex: this is checking that the market was disputed, not that the user has stake.

epheph: Forgot we left-joined! 👍

PR #361, opened 2018-05-16
relax contraint to allow 0 value transfers which do happen internally

nuevoalex:


nuevoalex:
While not strictly related to the branch title I added a fix I had to do locally to a getter as well. The escape hatches getAllOrders was returning stale order ids.

PR #365, opened 2018-05-17
changes to aws deploy script

petong:
deploy changes for stable


scripts/release-docker.sh
```diff
@@ -0,0 +1,44 @@
...
+    stable)
+        network="stable"
+        cluster="stable-augur-net"
+        ;;
+    *)
+        network=${augur_env}
+        ;;
+esac
+
+docker build . --build-arg ethereum_network=${network} --tag augurproject/augur:${augur_env} --tag augurproject/augur:$version || exit 1
```
epheph: Necessary when we're in `set -e` ?

pgebheim: I think with set -e that's rendundant

PR #377, opened 2018-05-18
Pgebheim/ch10359/get user trading history is returning too

pgebheim:


JohnDanz:
PL is messed up I think, but trading works and that's what this is here to fix first and foremost.

JohnDanz:
waiting on Travis, PL fixed.

PR #384, opened 2018-05-20
queryModifier often tries to sort by things that it shouldn't

pgebheim:
Story details: https://app.clubhouse.io/augur/story/10203

I've broken down sorting into two cases:

For fields which we auto-select as BN (using post-process-database-results.ts) -- we will sort in userland, and manually apply offset + limit;

For fields which are numerically sortable in the DB, we use a DB sort.


src/server/getters/database.ts
```diff
@@ -25,13 +28,78 @@ export interface Dictionary {
...
+  let sortDescending: boolean = defaultSortOrder.toLowerCase() === "desc";
+
+  if (sortBy != null) {
+    sortField = sortBy;
+    if (typeof(isSortDescending) !== "undefined" && isSortDescending !== null) {
+      sortDescending = isSortDescending;
+    }
+  }
+
+  query.select(db.raw(`?? as "xMySorterFieldx"`, [sortField])).asCallback((error: Error|null, rows: Array<RowWithSort>) => {
```
epheph: double `??`

src/server/getters/database.ts
```diff
@@ -25,13 +28,78 @@ export interface Dictionary {
...
+  callback: GenericCallback<Array<T>>,
+): void {
+  type RowWithSort = T & {xMySorterFieldx: BigNumber};
+
+  let sortField: string = defaultSortBy;
+  let sortDescending: boolean = defaultSortOrder.toLowerCase() === "desc";
+
+  if (sortBy != null) {
+    sortField = sortBy;
+    if (typeof(isSortDescending) !== "undefined" && isSortDescending !== null) {
```
epheph: what if they passed in `isSortDescending: false` ?

pgebheim: If so, the. The nullity and undefined checks will pass and sortDescending fill be set to false on line 65.

src/server/getters/database.ts
epheph: ahh cool!

src/server/getters/database.ts
epheph: oop, see it now!

PR #385, opened 2018-05-20
initial reporters and crowdsourcers of forking market need to show

nuevoalex:


src/server/getters/get-reporting-fees.ts
```diff
@@ -288,6 +288,7 @@ function getStakedRepResults(db: Knex, reporter: Address, universe: Address, cal
 
     const marketDisputed: AddressMap = {};
 
+    console.log(result.crowdsourcers);
```
epheph: meant to keep this?

PR #386, opened 2018-05-20
Epheph/ch11260/handle states transitions for non forked

epheph:


epheph:
Store if a universe is forked on the universe row (instead of having to look across all markets to see if there is one).
This is part 1 of the referenced tickert

PR #391, opened 2018-05-21
Epheph/ch11260/handle states transitions for non forked

epheph:
This PR addresses a few things:
1.) Markets in a forked market were progressing. Like if one was AWAITING_WINDOW, it would hit CROWDSOURCING like normal, when it isn't true. Same of some other states
2.) Markets coming out of initial reporting were always AWAITING_WINDOW, but they should be AWAITING_FORK_MIGRATION
3.) Crowdsourcers were being left completed in the forked uni and disputeRounds value not being updated.

### DIFF
--- a/src/blockchain/log-processors/initial-report-submitted.ts
+++ b/src/blockchain/log-processors/initial-report-submitted.ts
@@ -9,7 +9,7 @@ export function processInitialReportSubmittedLog(db: Knex, augur: Augur, log: Fo
   db("universes").first("forked").where({ universe: log.universe }).asCallback((err, universeRow?: { forked: boolean }) => {
     if (err) return callback(err);
     if (universeRow == null) return callback(new Error("No universe in initial report"));
-    const marketState = universeRow.forked ? ReportingState.AWAITING_FORK_MIGRATION ? augur.constants.REPORTING_STATE.AWAITING_NEXT_WINDOW;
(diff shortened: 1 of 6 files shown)


bthaile:
How do markets in `AWAITING_FORK_MIGRATION` leave this state? I would think after they get migrated to the new Universe. Then they would go to `AWAITING_NEXT_WINDOW`. I'm not seeing that behavior. The market is kept in `AWAITING_FORK_MIGRATION`. 

Need more information on what happens to markets in `AWAITING_FORK_MIGRATION`.

bthaile:
I don't think the UI needs to know about `AWAITING_FORK_MIGRATION`, IMO the market should move to `AWAITING_NEXT_WINDOW` after it gets migrated.

epheph:
"Also, I would think the markets would go to round 0, being it's like they just got initially reported on." that should be taken care of in this pr. was it?

bthaile:
I think too many markets got moved to `AWAITING_NEXT_WINDOW` markets that were in`OPEN_REPORTING` also got moved to `AWAITING_NEXT_WINDOW`, which isn't correct b/c they haven't been reported on.

epheph:
@bthaile you are right! Same thing applies to PRE_REPORTING and DESIGNATED_REPORTING, too. Basically, anything not awaiting_fork_migration.

Fixed.

bthaile:
It works well, migrating markets they end up in the right spot/state. approving.

PR #397, opened 2018-05-22
Achapman/ch11405/claiming winning shares doesn't update market

nuevoalex:


src/blockchain/log-processors/trading-proceeds-claimed.ts
```diff
@@ -18,14 +19,20 @@ export function processTradingProceedsClaimedLog(db: Knex, augur: Augur, log: Fo
   db("trading_proceeds").insert(tradingProceedsToInsert).asCallback((err?: Error|null) => {
     if (err) return callback(err);
     augurEmitter.emit("TradingProceedsClaimed", log);
-    callback(null);
+    refreshPositionInMarket(db, augur, log.market, log.sender, (err: Error|null) => {
```
epheph: Could just pass callback in to refreshPositionInMarket, if we aren't going to do anything in callback? (although putting above emit inside seems more correct)

nuevoalex: Went with simplified callbacks because less code

PR #399, opened 2018-05-22
Handle incomplete crowdsourcers correctly.

epheph:
We were using the crowdsourcer's fixed size to do calculations, but if a crowdsourcer is not complete, that isn't correct. A user who has 0.2 in a 0.2 crowdsourcer that is incomplete is owed 100% of eth fees for it.


epheph:
Adding crowdsourcer as a token ended up not being the way I went, but I like having it in there anyway

epheph:
Don't merge

epheph:
@nuevoalex i also just threw in there a commit for not bothering to look at balance = 0 ,which was causing the redemption on redeemed crowdsourcers

PR #400, opened 2018-05-22
Use REP balance instead of making call to universe, which can change

epheph:


epheph:
I'll prolly change when this happens:
https://github.com/AugurProject/augur-core/pull/644

PR #407, opened 2018-05-30
Update volumetrics without using off-chain calls for, required for consistency

epheph:


epheph:
This required extensive test data change because we had bad test data. `0x1000000000000000000000000000000000000000` was used for a sharetoken as well as a FeeWindow/participationToken . Split them out and updated tests accordingly

PR #413, opened 2018-05-31
version bump

nuevoalex:
old build was bad :[


nuevoalex:
Bad linked libraries. May not have affected this as much as augur-ui but I'm rebuilding everything to be sure

PR #447, opened 2018-06-14
Epheph/ch12476/return filled orders in get orders

JohnDanz:
apparently this wasn't actually PR'ed so doing that


epheph:
I'd like to add a test to this specific case, but i'd like to refactor the off-chain call out of the processing, so i'll deal with in a different PR

PR #459, opened 2018-06-20
add looking in orders table for same transactionHash as token transfer

bthaile:
add looking for transaction hash in orders table as well as trades.


epheph:
The logic here is good, consider orders that have not been filled (and have not showed up in trades table yet).

PR #461, opened 2018-06-21
Tag augur-node docker image with core version

justinbarry:


epheph:
Interesting, so there's a different docker build system in /scripts/docker/(build|push).sh that pushes an augur-core tagged version. Where is this docker:release run? I'd be perfectly happy to delete the /scripts/docker one, as it isn't tied to any CI system

justinbarry:
@epheph Those scripts are relatively recent. Should we be using this instead?

justinbarry:
^ Moved discussion to our ops channel.

PR #493, opened 2018-07-06
Petong/stable merge 201807060026

petong:


pgebheim:
Looks like the right set of changes, will double check why this branch isn't FFable after we merge.

PR #505, opened 2018-07-09
pass down error callback to pass errors to augur-app or caller instea…

bthaile:
…d of throwing from setInterval

https://app.clubhouse.io/augur/story/13483/augur-app-connection-popup-error-when-trying-to-connect-to-custom-second-time


src/controller.ts
```diff
@@ -42,7 +42,7 @@ export class AugurNodeController {
     this.running = true;
     this.errorCallback = errorCallback;
     try {
-      this.db = await createDbAndConnect(this.augur, this.networkConfig, this.databaseDir);
+      this.db = await createDbAndConnect(this.augur, this.networkConfig, errorCallback, this.databaseDir);
```
epheph: Can you move the callback to the beginning for consistency?

epheph: Or end?

src/blockchain/monitor-ethereum-node-health.ts
```diff
@@ -1,20 +1,29 @@
...
   }
-  monitorEthereumNodeHealthId = setInterval(() => {
-    augur.api.Universe.getController({ tx: { to: universe } }, (err: Error, universeController: string) => {
-      if (err) throw err;
-      if (universeController !== controller) {
-        throw new Error(`Controller mismatch. Configured: ${controller} Found: ${universeController}`);
-      }
-    });
-  }, 5000);
+  try {
```
epheph: What error would be caught with try?

src/controller.ts
bthaile: can not be at the end because of option databaseDir param.

seems augur is passed in as first param as general convention.

src/blockchain/monitor-ethereum-node-health.ts
bthaile: agreed, removed it

src/controller.ts
bthaile: Will move the error callback to the beginning

PR #535, opened 2018-07-12
this gives ability to remove log and set by augur-app, needed to remo…

bthaile:
…ve double logging

https://app.clubhouse.io/augur/story/13794/augur-node-logging-logs-empty-messages

this will remove double logging. empty entries are coming from augur-app electron-log.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -112,6 +112,10 @@ export class AugurNodeController {
     this.logger.addLogger(logger);
   }
 
+  public clearLoggers() {
+    this.logger.clear();
+  }
+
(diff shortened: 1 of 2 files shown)


src/controller.ts
```diff
@@ -112,6 +112,10 @@ export class AugurNodeController {
     this.logger.addLogger(logger);
   }
 
+  public clearLoggers() {
+    this.logger.clear();
```
justinbarry: Should that be `this.logger.clearLoggers()`?

bthaile: .clear method already existed so gtg.

PR #537, opened 2018-07-13
Change getContractAddresses to getSyncData, return block info

epheph:


bthaile:
Nice they aren't breaking changes.

PR #541, opened 2018-07-13
need to catch websocket errors

epheph:


epheph:
https://github.com/AugurProject/augur-app/issues/131

PR #553, opened 2018-07-19
account for price correctly when orphaning

nuevoalex:
Forgot to account for the actual price in the original implementation. The log removal could be simplified too since we can always know that the last log is not orphaned .


nuevoalex:
Found more issues testing in Prod. Apologies on the back and forth. Testing this takes a long time...

PR #554, opened 2018-07-19
johndanz/ch14144/group-categories-with-same-name-but-different

JohnDanz:
This is the patch to make sure that Categories are saved and returned as ALL UPPER CASE VALUES so that there aren't multiple versions of the `SPORTS` category for example. Currently if the user creates a market with the category as `sports` or `Sports` or `SPORTS` that will produce 3 different categories in the UI that are all labeled as `SPORTS` and it becomes confusing/fragments popularity numbers for the same topics.

https://app.clubhouse.io/augur/story/14144/group-categories-with-same-name-but-different-cases-together-in-augur-node


src/server/getters/get-categories.ts
```diff
@@ -12,6 +12,6 @@ export function getCategories(db: Knex, universe: Address, sortBy: string|null|u
   queryModifier(db, query, "popularity", "desc", sortBy, isSortDescending, limit, offset, (err: Error|null, categoriesInfo?: Array<CategoriesRow>): void => {
     if (err) return callback(err);
     if (!categoriesInfo) return callback(null);
-    callback(null, categoriesInfo.map((categoryInfo: CategoriesRow): CategoriesRow => Object.assign(categoryInfo, { popularity: categoryInfo.popularity.toString() })));
+    callback(null, categoriesInfo.map((categoryInfo: CategoriesRow): CategoriesRow => Object.assign(categoryInfo, { popularity: categoryInfo.popularity.toString(), category: categoryInfo.category.toUpperCase() })));
```
epheph: Can we get rid of the object assign since we're explicitly setting both keys  in a later object?

epheph:
You hinted at this in a different conversation, but this will have different behavior if coming from an existing DB vs sync'ing new, because it isn't rolling up the existing data for categories. It's probably not a huge deal, and we might require a resync again soon? something to consider.

src/blockchain/log-processors/market-migrated.ts
```diff
@@ -41,7 +41,7 @@ export function processMarketMigratedLog(db: Knex, augur: Augur, log: FormattedE
             db.select("popularity").from("categories").where({ category, universe: log.newUniverse }).asCallback((err: Error|null, categoriesRows?: Array<CategoriesRow>): void => {
               if (err) return callback(err);
               if (categoriesRows && categoriesRows.length) return callback(null);
-              db.insert({ category, universe: log.newUniverse }).into("categories").asCallback(callback);
+              db.insert({ category: category.toUpperCase(), universe: log.newUniverse }).into("categories").asCallback(callback);
```
epheph: shouldn't be necessary, since the value is coming from the DB, but def doesn't hurt, and no reason to remove it

JohnDanz: yeah the goal was to just try and make sure, (also a attempt to potentially make it backward compatible, but based on our convo idk that it will be and we will probably need a full resync)

PR #559, opened 2018-07-20
allow resume from bulk sync if interrupted

epheph:


bthaile:
npm run start
Or to run from a clean db. npm run clean-start

epheph:
@steveklebanoff you are on this branch and have run `npm run build` ? It works for me, the "fromBlock" is different when the process is killed mid-sync (so long as there has been at least one "new block" log spit from first run)

PR #560, opened 2018-07-21
Pgebheim/ch14218/remove positions caching

pgebheim:
Remove off-chain calls for positions and base the results of getUserTradingPosition on getProfitLos.

8 off-chain calls remain to get the lastPrice for getMarketsInfo until this is fixed.


pgebheim:
Slotting this in for 1.0.6

PR #566, opened 2018-07-24
Epheph/ch11769/enable ability to search all markets even

bthaile:
https://app.clubhouse.io/augur/story/11769/enable-ability-to-search-all-markets-even-if-they-are-not-already-loaded


src/blockchain/log-processors/market-created.ts
```diff
@@ -100,6 +105,9 @@ export function processMarketCreatedLog(db: Knex, augur: Augur, log: FormattedEv
             (next: AsyncCallback): void => {
               db.insert(marketsDataToInsert).into("markets").asCallback(next);
             },
+            (next: AsyncCallback): void => {
+              db.raw("insert into search_en(marketId, content) values( ?, ? )", [fullTextStringInsert.marketId, fullTextStringInsert.content]).asCallback(next);
```
epheph: Does the normal db.insert({}) method not work for that virtual table?

bthaile: I’d need to try that. I went with raw b/c it worked

src/utils/content-search-builder.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (marketsDataToInsert === null) throw new Error("market insert row object can't be null");
+
+  const content = [
+    marketsDataToInsert.marketId,
+    marketsDataToInsert.category,
+    marketsDataToInsert.shortDescription !== null ? marketsDataToInsert.shortDescription : "",
+    marketsDataToInsert.tag1 !== null ? marketsDataToInsert.tag1 : "",
+    marketsDataToInsert.tag2 !== null ? marketsDataToInsert.tag2 : "",
+    marketsDataToInsert.resolutionSource !== null ? marketsDataToInsert.resolutionSource : "",
+    marketsDataToInsert.scalarDenomination != null ? marketsDataToInsert.scalarDenomination : "",
```
epheph: I wonder if outcome names would be interesting to add? could be a future improvement

src/server/getters/get-markets-search.ts
```diff
@@ -0,0 +1,21 @@
+import * as Knex from "knex";
+import { Address, MarketsContractAddressRow } from "../../types";
+import { queryModifier, getMarketsWithReportingState } from "./database";
+
+// Returning marketIds should likely be more generalized, since it is a single line change for most getters (awaiting reporting, by user, etc)
+export function getMarketsSearch(db: Knex, universe: Address, search: string, sortBy: string|null|undefined, isSortDescending: boolean|null|undefined, limit: number|null|undefined, offset: number|null|undefined, callback: (err?: Error|null, result?: any) => void): void {
```
epheph: Can we add a type for `result?: any`?

bthaile: I agree, for future expansion.

src/utils/content-search-builder.ts
bthaile: I think only categorical markets use description, I think we should be good.

src/server/getters/get-markets-search.ts
bthaile: changed to return array of strings.

PR #573, opened 2018-07-27
use env variables to pass in lag and retries variables to augur node …

bthaile:
…controller

https://app.clubhouse.io/augur/story/14059/allow-for-propagationdelaywaitmillis-to-passed-in-via-command-line-to-run-augur-node

### DIFF
--- a/src/runServer.ts
+++ b/src/runServer.ts
@@ -28,12 +28,14 @@ function start(retries: number, config: any, databaseDir: any) {
     logger.warn("Reconnect to Ethereum node");
   });
 
-  function errorCatch(err: any) {
+  function errorCatch(err: Error) {
     if (retries > 0) {
+      logger.warn(err.message);
       retries--;
       augurNodeController.shutdown();
-      setTimeout(start(retries, config, databaseDir), 1000);
+      setTimeout(() => start(retries, config, databaseDir), 1000);
     } else {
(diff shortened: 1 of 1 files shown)


src/runServer.ts
```diff
@@ -5,22 +5,39 @@ import { logger } from "./utils/logger";
...
-const augurNodeController = new AugurNodeController(augur, networkConfig, databaseDir);
+let config = networkConfig;
+if (maxRetries) config = Object.assign({}, config, { maxRetries });
+if (propagationDelayWaitMillis) config = Object.assign({}, config, { propagationDelayWaitMillis });
+const retries: number = parseInt(maxRetries, 10);
+
+function start(retries: number, augur: Augur, config: any, databaseDir: any) {
+  const augurNodeController = new AugurNodeController(augur, config, databaseDir);
+
+  augur.rpc.setDebugOptions({ broadcast: false });
```
epheph: What's this debug about?

epheph:
I think we should make a different variable for the global retry vs. the eth_call. The "maxRetries" you pass into network options is how many times, internally, it will retry an `eth_call` before actually calling back an error. The outer is how many times the whole thing can fail catastrophically. I could imagine being on a very reliable eth node and never wanting to retry an eth_call, but have the whole system retry if there's a more general error

PR #574, opened 2018-07-27
Speed up block details download

epheph:
We use promises to download block details and start that process on receipt of a batch of logs, not when those logs are processed (which could be much later and happens inside of a transaction while everything else is stopped).

In this PR we also fix the "double transaction" that occurs during bulk sync

### DIFF
--- a/src/blockchain/download-augur-logs.ts
+++ b/src/blockchain/download-augur-logs.ts
@@ -14,9 +14,14 @@ interface BlockDetailsByBlock {
   [blockNumber: number]: BlockDetail;
 }
 
+function extractBlockNumbers(batchOfAugurLogs: Array<FormattedEventLog>) {
+  return _.uniq(batchOfAugurLogs.map((augurLog) => augurLog.blockNumber));
+}
+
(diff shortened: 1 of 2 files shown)


src/blockchain/download-augur-logs.ts
```diff
@@ -70,23 +76,29 @@ export function downloadAugurLogs(db: Knex, augur: Augur, fromBlock: number, toB
...
-    }));
-  }, (err) => {
-    if (!batchLogProcessQueue.paused) {
-      batchLogProcessQueue.push(() => {
-        callback(err);
-        batchLogProcessQueue.kill();
+      if (!batchOfAugurLogs || batchLogProcessQueue.paused) return;
+      const blockNumbers = extractBlockNumbers(batchOfAugurLogs);
+      const blockDetailPromise = lastBlockDetails.then(() => fetchAllBlockDetails(augur, blockNumbers));
+      lastBlockDetails = blockDetailPromise;
```
nuevoalex: why is this set here? Doesn't look like its used

PR #581, opened 2018-07-30
Emitter to use an enum of event names

epheph:
So that we can check during subscription time if the event being subscribed to exists or not, and return an error if it does not.

This code is no-op, it just changes internal structure to start using the enum everywhere. I'll throw the actual improvement in a different PR so the change is more obvious


src/blockchain/log-processors/database.ts
```diff
@@ -68,7 +69,7 @@ export function updateActiveFeeWindows(db: Knex, blockNumber: number, timestamp:
                     }
                     if (newActiveFeeWindowRows != null) {
                       newActiveFeeWindowRows.forEach((newActiveFeeWindowRow) => {
-                        augurEmitter.emit("FeeWindowOpened", Object.assign({
+                        augurEmitter.emit(SubscriptionEventNames.FeeWindowClosed, Object.assign({
```
adrake33: Shouldn't this be `SubscriptionEventNames. FeeWindowOpened `?

PR #589, opened 2018-08-02
Add digital ocean run script for docker machine

pgebheim:


pgebheim:
This was adapted from https://github.com/AugurProject/augur-node/pull/295

PR #591, opened 2018-08-05
Helmet patch

joeykrug:


epheph:
I wonder how much of the helmet system applies to websockets, where most of the data is available? It looks like some of it, but perhaps not all? (the headers are only on the initial connect). We do have HTTP server as well for status endpoints. I say merge it, but let's see how much it actually improves websocket security

PR #609, opened 2018-08-21
Update unit test require paths to use source files

justinbarry:


pgebheim:
We should change all of the test files to be .ts so that ts-node builds it all before running. If not it will stall to JIT any new ts it encounters and the test will time out.

epheph:
Yep, that's what @justinbarry and i discussed. He has a branch where he started that work

pgebheim:
@justinbarry @epheph -- Is this closable?

epheph:
@justinbarry if we want to really do this, we need to basically halt development/tests improvement and get it done, yea?

justinbarry:
I'm still unclear why the timeout is an unacceptable option. But we can having a separate tsconfig file for the tests. See: https://stackoverflow.com/questions/35470511/setting-up-tsconfig-with-spec-test-folder <https://stackoverflow.com/questions/35470511/setting-up-tsconfig-with-spec-test-folder>

epheph:
I don't think it was ever dismissed as an unacceptable option. It would work and, when we eventually get tests in TS, we can get rid of it.

justinbarry:
It we can go that route then we can merge this and start migrating tests piecemeal.

epheph:
OK, i'll clean up the conflict

justinbarry:
Sweet!

PR #620, opened 2018-08-29
WIP: Postgres + Heroku Support

pgebheim:


Procfile
```diff
@@ -1 +1 @@
-web: npm start
+web: npm run migrate_pg && npm start -- $NETWORK
```
epheph: is pre-migrating necessary? migration is handled inside AN now during runtime

src/migrations/20171005123242_orders.ts
```diff
@@ -3,7 +3,7 @@ import * as Knex from "knex";
 exports.up = async (knex: Knex): Promise<any> => {
   return knex.schema.dropTableIfExists("orders").then(async (): Promise<any> => {
     return knex.schema.createTable("orders", (table: Knex.CreateTableBuilder): void => {
-      table.string("orderId", 42).primary().notNullable();
```
epheph: Huh, so sqlite let us get away with storing too much in that column

src/migrations/20171005123242_orders.ts
```diff
@@ -18,11 +18,13 @@ exports.up = async (knex: Knex): Promise<any> => {
...
-      table.specificType("originalAmount", "varchar(255) NOT NULL CONSTRAINT nonnegativeAmount CHECK (ltrim(\"amount\", '-') = \"amount\")");
+      // Originally this constraint was called nonnegativeAmount,
+      // but it is not allowed in Postgres to have two constraints with the same name.
+      // So the constraint is renamed but for SQLite3 it is the same to keep compatibility.
+      table.specificType("originalAmount", `varchar(255) NOT NULL CONSTRAINT ${knex.client.config.client === "sqlite3" ? "nonnegativeAmount" : "nonnegativeOriginalAmount"} CHECK (ltrim(\"amount\", '-') = \"amount\")`);
       table.specificType("tokensEscrowed", "varchar(255) NOT NULL CONSTRAINT nonnegativeTokensEscrowed CHECK (ltrim(\"tokensEscrowed\", '-') = \"tokensEscrowed\")");
       table.specificType("sharesEscrowed", "varchar(255) NOT NULL CONSTRAINT nonnegativeSharesEscrowed CHECK (ltrim(\"sharesEscrowed\", '-') = \"sharesEscrowed\")");
-      table.boolean("orphaned").defaultTo(0);
-      table.string("tradeGroupId", 42);
+      table.string("tradeGroupId", 66);
```
epheph: orphaned removed?

src/migrations/20171005123329_transfers.ts
```diff
@@ -4,7 +4,7 @@ exports.up = async (knex: Knex): Promise<any> => {
   return knex.schema.dropTableIfExists("transfers").then((): PromiseLike<any> => {
     return knex.schema.raw(`CREATE TABLE transfers (
```
epheph: it would be really nice to get rid of these .raw(CREATE)'s and turn them into proper create tables

Procfile
pgebheim: This is probably no longer necessary, indeed. I'll remove it.

src/migrations/20171005123242_orders.ts
pgebheim: woop, that was a mis-merge! Good batch

src/migrations/20171005123242_orders.ts
pgebheim: Yeah, SQLite doesn't really pay attention to the types you give it. It will store whatever you shove at it in the column, its kinda weird.

src/migrations/20171005123329_transfers.ts
pgebheim: Agreed --- maybe another lager cleanup since this patch is kinda unwieldy anyway?

PR #651, opened 2018-09-14
Github: Parity Node: Could not find balance for token decrease

epheph:
Story details: https://app.clubhouse.io/augur/story/15116

### DIFF
--- a/src/blockchain/download-augur-logs.ts
+++ b/src/blockchain/download-augur-logs.ts
@@ -1,4 +1,4 @@
-import { Augur } from "augur.js";
+import { Augur, BlockRange } from "augur.js";
 import { eachSeries, mapLimit, queue } from "async";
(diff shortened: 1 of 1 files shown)


epheph:
"If no logs were included in this batch, grab the first, middle, and last block number and fetch those details. If we don't receive any block details, but also don't receive an error, inform the user they likely have a partially-sync'd warp node

justinbarry:
@epheph I'm getting an error when building:
`Error:(1, 17) TS2305: Module '"augur-node/node_modules/augur.js/index"' has no exported member 'BlockRange'.`

pgebheim:
Ok we'll need to add that type import I think

justinbarry:
This will never pass w/o an augur.js release.

justinbarry:
Was merged in as part of #660

PR #654, opened 2018-09-14
fix migration script for volume

nuevoalex:


justinbarry:
Just so we are clear. This migration hasn’t been out in the wild has it?

PR #665, opened 2018-10-01
Pgebheim/ch17134/token supply row not removed in market created

pgebheim:


test/unit/blockchain/log-processors/market-created.js
```diff
@@ -1,4 +1,4 @@
-"use strict";
+"utse strict";
```
epheph: utse

src/blockchain/log-processors/market-created.ts
```diff
@@ -141,26 +141,25 @@ export function processMarketCreatedLog(db: Knex, augur: Augur, log: FormattedEv
...
-    (next: AsyncCallback): void => {
-      db.from("market_state").where({ marketId: log.market }).del().asCallback(next);
-    },
-    (next: AsyncCallback): void => {
-      db.from("search_en").where({ marketId: log.market }).del().asCallback(next);
-    },
-  ], (err) => {
-    if (err) callback(err);
+  marketCreatedLogRemoval(db, augur, log).then(() => {
     augurEmitter.emit(SubscriptionEventNames.MarketCreated, log);
```
epheph: emitting here works for market log removal, but in other places, we add data from the context of the processor, so it wouldn't be available here. i wonder if we should make this a responsibility of the promise (and then, once they are all pormises, just have the log-processor use the promise directly.

pgebheim: Like we resolve the promise with the transformed data?

epheph: yea i think in the end, we the function that calls into the blockchain processor take it as a promise directly (once they are all promises) so we don't have to have a shim everywhere

test/unit/blockchain/log-processors/market-created.js
pgebheim: Oh boy! How did that happen ;-)

src/blockchain/log-processors/market-created.ts
pgebheim: I like that.

pgebheim: Random thought -- can we use Typescript generics to kinda give us the ability to export either a function-with-callback or a async-function??

epheph: Yes, but it will still come down to:
```
if isCallbackStyle(processor) processor(...)
else isPromiseStyle(processor) processor(...)

PR #683, opened 2018-10-16
Re-order off-chain calls to top of processing

epheph:
Fix updateMarketFeewindow to only hit chain if necessary


src/blockchain/log-processors/market-created.ts
```diff
@@ -44,111 +68,109 @@ export function processMarketCreatedLog(db: Knex, augur: Augur, log: FormattedEv
...
-        const shareTokens = new Array(numOutcomes);
-        forEachOf(shareTokens, (_: null, outcome: number, nextOutcome: ErrorCallback): void => {
-          augur.api.Market.getShareToken(Object.assign({ _outcome: outcome }, marketPayload), (err: Error|null, shareToken?: Address): void => {
-            if (err) return nextOutcome(err);
-            shareTokens[outcome] = shareToken;
-            nextOutcome(null);
-          });
-        }, (err: Error|null): void => {
+
+        series([
```
pgebheim: Looks like this `series` should be able to be pulled out into just a sequential list of awaits pretty easily, add to the list?

PR #700, opened 2018-11-08
remove case that undermines the actual check being done

nuevoalex:
So for context this was resulting in an error during blockstream which would halt the stream but allow AN to continue running. That seems like a larger issue that should be addressed at some point. The visible results were that AN wasn't updating (obv) and other strange behavior in the UI like no time data being available (the original bug this issue was for).

What was occurring was that in the case the user was up to date but within the handoff block window this code would simply make them fetch just the block they last synced +1. This effectively gets rid of any handoff block window and also allows the sync to start at a block that doesn't yet exist in the case where the user's latest synced block is also the highest block. The later case is what was causing the error  that was manifesting where AN stalls and the UI shows weird data.

Since the handoff blocks number was basically being ignored previously I reduced it back to 5.

### DIFF
--- a/src/blockchain/bulk-sync-augur-node-with-blockchain.ts
+++ b/src/blockchain/bulk-sync-augur-node-with-blockchain.ts
@@ -34,7 +34,7 @@ export async function bulkSyncAugurNodeWithBlockchain(db: Knex, augur: Augur): P
   } else {
     fromBlock = lastSyncBlockNumber == null ? uploadBlockNumber : lastSyncBlockNumber + 1;
   }
-  let handoffBlockNumber = highestBlockNumber - BLOCKSTREAM_HANDOFF_BLOCKS;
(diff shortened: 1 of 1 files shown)


src/blockchain/bulk-sync-augur-node-with-blockchain.ts
```diff
@@ -6,7 +6,7 @@ import { augurEmitter } from "../events";
 import { logger } from "../utils/logger";
 import { SubscriptionEventNames } from "../constants";
 
-const BLOCKSTREAM_HANDOFF_BLOCKS = 105;
+const BLOCKSTREAM_HANDOFF_BLOCKS = 5;
```
epheph: this value of 105 slipped through an earlier PR 😢 I use it for testing

epheph:
This whole section is a patch over "what do we do if there is a re-org beyond where the sync started", such as storing raw logs in DB. This bug has always been there, but 105 slipped through caused it to be commonly hit.

PR #705, opened 2018-11-10
add bulksSync column and record tx hashes

nuevoalex:
adding some information useful when debugging

### DIFF
--- a/src/blockchain/download-augur-logs.ts
+++ b/src/blockchain/download-augur-logs.ts
@@ -68,7 +68,7 @@ async function processBatchOfLogs(db: Knex, augur: Augur, allAugurLogs: Array<Fo
     });
     const dbWriteFunctions = await Promise.all(dbWritePromises);
     await db.transaction(async (trx: Knex.Transaction) => {
-      await processBlockByBlockDetails(trx, augur, blockDetailsByBlock[blockNumber]);
+      await processBlockByBlockDetails(trx, augur, blockDetailsByBlock[blockNumber], true);
       logger.info(`Processing ${dbWriteFunctions.length} logs`);
       for (const dbWriteFunction of dbWriteFunctions) {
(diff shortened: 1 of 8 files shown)


src/blockchain/download-augur-logs.ts
```diff
@@ -68,7 +68,7 @@ async function processBatchOfLogs(db: Knex, augur: Augur, allAugurLogs: Array<Fo
     });
     const dbWriteFunctions = await Promise.all(dbWritePromises);
     await db.transaction(async (trx: Knex.Transaction) => {
-      await processBlockByBlockDetails(trx, augur, blockDetailsByBlock[blockNumber]);
+      await processBlockByBlockDetails(trx, augur, blockDetailsByBlock[blockNumber], true);
       logger.info(`Processing ${dbWriteFunctions.length} logs`);
       for (const dbWriteFunction of dbWriteFunctions) {
```
epheph: transactionHashes not being inserted during bulk load, i'll fix in a different PR

src/migrations/20181109151249_transactionHashes.ts
```diff
@@ -0,0 +1,15 @@
+import * as Knex from "knex";
+
+exports.up = async (knex: Knex): Promise<any> => {
+  return knex.schema.dropTableIfExists("transactionHashes").then((): PromiseLike<any> => {
+    return knex.schema.raw(`CREATE TABLE transactionHashes (
+      "transactionHash" varchar(66) PRIMARY KEY NOT NULL,
```
epheph: This shouldn't be primary key. on removal, the removed flag is set, but the same transaction is likely to come back, creating a duplicate primary key issue. then when it comes back, we don't have information about when the new tx showed up, so our data is incomplete. I'll remove this, add blockHash, and create a unique index on blockHash/transactionHash.

PR #709, opened 2018-11-16
Warn if we detect bad light-mode getLogs response

epheph:


pgebheim:
I love good error messages.

Issue #714, opened 2018-11-19
Error when syncing with Postgres


mattxsmith:
When syncing with Postgres, the new table _transactionHashes_ is causing an error:

`select * from "transactionHashes" where "transactionHash" = $1 - relation "transactionHashes" does not exist`

I realized this is an error due to the way Postgres is capitalizing the table name. When syncing, the query is looking for a table with a capital H: _transactionHashes_. When you look at the actual table created, there is no capitalization: _transactionhashes_. 

This can be fixed by adding double quotes around _transactionHashes_ on line 5 in src/migrations/20181109151249_transactionHashes.ts.

adrake33:
Created a ticket for this in our internal tracking system: https://app.clubhouse.io/augur/story/18149/github-error-when-syncing-with-postgres

adrake33:
Closing this since a fix has been merged.

PR #718, opened 2018-11-20
Add double quotes to CREATE TABLE statement

adrake33:
GH issue: https://github.com/AugurProject/augur-node/issues/714

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "augur-node",
-  "version": "7.0.1",
+  "version": "7.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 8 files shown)


epheph:
This is an improvement, but even better is using the native knex schema functional format
👍

adrake33:
@epheph Ah, ok. So would the `return knex.schema.raw` line look like this instead:

```
return knex.schema.createTable("transactionHashes", (table: Knex.CreateTableBuilder): void => {
      table.string("transactionHash", 66).primary().notNullable();
      table.integer("blockNumber").notNullable();
      table.boolean("orphaned").defaultTo(0);
    });
```

epheph:
Yas. Looks really nice.

And i think we've run into that whole boolean/0 thing before and thought the way you did it there is correct?

adrake33:
I think so? It's done that way in the orders.ts file.

epheph:
👍 ship it!

PR #723, opened 2018-11-21
Add return

adrake33:


epheph:
Yup, that's important; we've had races with that otherwise.

PR #744, opened 2018-12-05
Add an version to block sync data

epheph:
@JohnDanz 's PR, I just clicked "Create"


bthaile:
I'm not getting the version from getSyncData(), I would expect.
 
```  "netId": "1",
  "net_version": "1",
  "version": "5.9.0-1"

```

I thought I would get the version in package.json which is 7.1.5

bthaile:
lot of versions going on here. `version` is for augur.js. Does that need to be more explicit. minor nitpick
```
  "augurNodeVersion": "7.1.5",
  "netId": "102",
  "net_version": "102",
  "version": "6.1.2"
```

PR #748, opened 2018-12-18
handle a mid start shutdown without crashing

nuevoalex:


src/controller.ts
```diff
@@ -49,6 +49,8 @@ export class AugurNodeController {
       const handoffBlockNumber = await bulkSyncAugurNodeWithBlockchain(this.db, this.augur);
       this.controlEmitter.emit(ControlMessageType.BulkSyncFinished);
       this.logger.info("Bulk sync with blockchain complete.");
+      // We received a shutdown so just return.
+      if (!this.running || this.db == null) return;
```
pgebheim: There's a helper in here for this, you can do:

```
if (!this.isRunning()) ...
```

I'll just update ;-)

PR #751, opened 2018-12-20
make blocks per chunck configurable and move orphan check to after bu…

nuevoalex:
…lk sync / block processing


src/blockchain/log-processors/order-created.ts
```diff
@@ -70,46 +70,13 @@ export async function processOrderCreatedLogRemoval(augur: Augur, log: Formatted
...
-    const orderIds: Array<string> = await getExistingOrders(requestData);
-
-    // Erroring here is expected in the case where we have more orders than are supported by the call used. We correct at some future order creation which must occur for there to be more orders now
-    return db.from("orders").whereNotIn("orderId", orderIds).where(queryData).update({ orphaned: true });
-  } catch (err) {
-    console.warn(`Failed to mark orphaned orders for ${orderData.marketId}`);
-  }
+  const result: { count: number } = await db.first(db.raw("count(*) as count")).from("pending_orphan_checks").where(pendingOrderData);
+  if (result.count > 0) return;
+  return await db.insert(pendingOrderData).into("pending_orphan_checks");
```
pgebheim: Nice touch storing that in the DB so we don't miss the check during a crash.

PR #754, opened 2019-01-02
Add warp sync

bthaile:
add "warp sync" feature

Every 100 blocks remove old warp sync file and create new one. Warp sync file is a zipped version of augur db. It is named by hash, network id, and db version.

Allow this file to be imported through controller, which will back up existing and replace with unzipped warp sync file.

added some formatting as well.


src/runServer.ts
```diff
@@ -25,9 +26,9 @@ if (maxRetries) config = Object.assign({}, config, { maxRetries });
 if (propagationDelayWaitMillis) config = Object.assign({}, config, { propagationDelayWaitMillis });
 const retries: number = parseInt(maxSystemRetries || "1", 10);
 
-function start(retries: number, config: any, databaseDir: any) {
```
epheph: Better type for isWarpSync here? (although i see they were already any before you modified this line)

src/setup/check-and-initialize-augur-db.ts
```diff
@@ -9,17 +9,12 @@ import { postProcessDatabaseResults } from "../server/post-process-database-resu
 import { monitorEthereumNodeHealth } from "../blockchain/monitor-ethereum-node-health";
 import { logger } from "../utils/logger";
 import { ConnectOptions, ErrorCallback } from "../types";
-
+import { DB_VERSION, DB_FILE } from "../constants";
```
epheph: 👍

src/sync/file-operations.ts
```diff
@@ -0,0 +1,74 @@
+import { format } from "util";
+import * as _ from "lodash";
+import * as fs from "fs";
+import * as path from "path";
+import * as zlib from "zlib";
+import { logger } from "../utils/logger";
+import { DB_WARP_SYNC_FILE_ENDING } from "../constants";
+
+export function getFileHash(filename: string): string {
+  const md5File = require("md5-file");
```
epheph: Can we use a proper import here?

src/sync/file-operations.ts
```diff
@@ -0,0 +1,74 @@
...
+      })
+      .on("finish", () => {
+        resolve();
+      });
+  });
+}
+
+export async function createWarpSyncFile(directoryDir: string, dbFileName: string, syncFilename: string): Promise<any> {
+  return new Promise<any>((resolve, reject) => {
+    fs.closeSync(fs.openSync(path.join(directoryDir, syncFilename), "w"));
```
epheph: "touch"ing the file? Is it necessary for the write stream? Would it be possible to do with promises insteqad of sync, since we're already in an sync function?

src/sync/file-operations.ts
```diff
@@ -0,0 +1,74 @@
...
+        reject(err);
+      })
+      .on("finish", () => {
+        resolve();
+      });
+  });
+}
+
+export async function createWarpSyncFile(directoryDir: string, dbFileName: string, syncFilename: string): Promise<any> {
+  return new Promise<any>((resolve, reject) => {
```
epheph: Make the Promise<void> since it will never return anything?

src/blockchain/process-block.ts
```diff
@@ -53,18 +54,33 @@ export async function processBlockAndLogs(db: Knex, augur: Augur, direction: Blo
...
+        .where({ blockHash: block.hash })
+        .del();
       // TODO: un-advance time
     }
   });
+  try {
+    if (isWarpSync && parseInt(block.number, 16) % DUMP_EVERY_BLOCKS === 0) {
+      // every X blocks export db to warp file.
+      const networkId: string = augur.rpc.getNetworkID();
+      BackupRestore.export(DB_FILE, networkId, DB_VERSION, DB_WARP_SYNC_FILE, databaseDir);
```
epheph: This should be await'd, so the file doesn't change while it is being exported

src/runServer.ts
epheph: Need to process into a boolean as well, since it's a string being covertly passed to something expecting a boolean.

src/sync/file-operations.ts
bthaile: The promise doesn't need to return anything, since there is file IO I wrapped it in a Promise.

epheph: hah! markdown destroyed my comment! Here's what it should have said:

```
Make the Promise<void> since it will never return anything?
```

src/runServer.ts
bthaile: fixed processing IS_WARP_SYNC env variable, typed the other parameters.

src/sync/file-operations.ts
bthaile: updated to use proper import.

src/sync/file-operations.ts
bthaile: removed touching file, not needed.

PR #761, opened 2019-01-07
Use blockchain timestamps for getPL

pgebheim:

PR #763, opened 2019-01-07
Update jest config

justinbarry:


pgebheim:
This will just be one change after #761 gets merged.

The change will effectively run JavaScript files through babel so we get hoisting for jest.mock

PR #777, opened 2019-01-11
Fix Volume Calculations

ryanberckmans:
Fixes AugurProject/augur#289

Ready for review. Ryan TODO - write final commit msg prior to rebase merge.


epheph:
We chatted about this, makes sense. The Math.min().multipledBy(numTicks) encompasses the sets that were destroyed for tokens. This will change future calculations, but there is no way to correct historical without re-processing. We decided it's probably just worth a release note.

PR #781, opened 2019-01-16
Fix 714

robert-e-davidson3:
My general approach to log ingestion is to provide tools for people to use to specify the database state before the actual tests begin. The main effect is that the complexity of the tests is now in the tests themselves instead of being hidden away

For this PR I made log ingestion optional. Some tests are updated to use log ingestion so they can serve as examples. All other updated tests are now explicitly seeding.


src/blockchain/download-augur-logs.ts
```diff
@@ -69,7 +70,9 @@ async function processBatchOfLogs(db: Knex, augur: Augur, allAugurLogs: Array<Fo
       await each(logs, async (log) => await insertTransactionHash(trx, blockNumber, log.transactionHash));
       logger.info(`Processing ${dbWriteFunctions.length} logs`);
       for (const dbWriteFunction of dbWriteFunctions) {
-        await dbWriteFunction(trx);
+        await dbWriteFunction(trx).catch((err) => {
```
epheph: Did you mean to leave this in? This will swallow the error and proceed

robert-e-davidson3: Oops yeah this should be removed. Thanks!

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
+  for (let p = 0; p < params.length; p++) {
+    const param = params[p];
+    if (!contains(args, param)) {
+      throw Error(`Required param "${param}" was not provided.`);
+    }
+  }
+}
+
+function seedDb(db) {
+  return Promise.resolve(null).then(async () => {
```
epheph: Could this also be represented as a native async function without the `return Promise.resolve(null).then(` ?

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
+    addresses: {
+      9090: {
+      },
+    },
+  };
+
+  return deepmerge(augur, additional || {});
+}
+
+const setupTestDb = (augur, logs, blockDetails) => {
```
epheph: Since all the other functions in this file are defined as standard functions, consider making this one consistent with the others?

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
+
+  const env = getEnv();
   const db = Knex(env);
   return db.migrate.latest(env.migrations)
-    .then(() => db.seed.run(env.seeds))
+    .then(() => {
+      const blockNumbers = Object.keys(blockDetails);
+      return processBatchOfLogs(db, augur, logs, blockNumbers, Promise.resolve(blockDetails));
+    })
+    .catch(err => console.error(err.stack))
```
epheph: Should we also re-throw here?

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
-module.exports = () => {
-  const env = Object.assign({}, environments.test, {
+
+function getEnv() {
+  return Object.assign({}, environments.test, {
     postProcessResponse: postProcessDatabaseResults,
   });
+}
+
+function contains(container, element) {
```
epheph: augur-node already pays the price of lodash for these sorts of functions, this one could be replaced by `_.includes`

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
+  for (let i = 0; i < container.length; i++) {
+    if (container[i] === element) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function required(args, ...params) {
+  args = Object.keys(args);
```
epheph: I think both functions could be replaced with:

```_.every(params, (param) => _.includes(args, param))```

File: test/unit/test.database.js (hunk 2 above)
Comment (robert-e-davidson3) [reply test/unit/test.database.js]: Yeah I can switch all of the promise stuff to async/await.

File: test/unit/test.database.js (hunk 3 above)
Comment (robert-e-davidson3) [reply test/unit/test.database.js]: Oh yeah good catch. I added the other functions but only modified this one so it slipped past me.

File: test/unit/test.database.js (hunk 4 above)
Comment (robert-e-davidson3) [reply test/unit/test.database.js]: I'll just remove it. It was helpful while I was figuring out how the logs worked but now it's probably not worth keeping.

File: test/unit/test.database.js (hunk 5 above)
Comment (robert-e-davidson3) [reply test/unit/test.database.js]: Oh perfect I'll switch to that.

File: test/unit/server/getters/get-winning-balance.js [hunk 7]
```diff
@@ -1,11 +1,11 @@
 const { BigNumber } = require("bignumber.js");
-const setupTestDb = require("../../test.database");
+const { setupTestDb, seedDb } = require("../../test.database");
 const { dispatchJsonRpcRequest } = require("src/server/dispatch-json-rpc-request");
 
 describe("server/getters/get-winning-balance", () => {
   let db;
   beforeEach(async () => {
-    db = await setupTestDb();
+    db = await setupTestDb().then(seedDb);
```
epheph: Is the `setupTestDb` ever run without the `seedDb` ? Wondering if it makes sense to do it in one place instead of in each getter? More a style thing than anything, the way you did it here might be more flexible

epheph: I see, so as you migrate to this system, you drop the `.then(seedDb)`, then only provide the log ingestion?

test/unit/test.database.js
```diff
@@ -1,13 +1,247 @@
...
+      }, args || {});
+    },
+  });
+}
+
+function hexify(integer) {
+  return "0x" + (integer).toString(16);
+}
+
+function makeFakeBlock(number, overrides) {
```
epheph: So there's really just two important bits here: blockNumber and timestamp. I see we could do
`makeFakeBlock(2, {timestamp: 4000});`, but it might be cool if timestamp was a first-class option (since it changes market state and dispute rounds and whatnot)? But this is totally workable, too.

test/unit/server/getters/get-winning-balance.js
robert-e-davidson3: Yup!

test/unit/test.database.js
```diff
@@ -1,13 +1,252 @@
...
+  augur = augur || new Augur();
+  logs = logs || [];
+  blockDetails = blockDetails || {};
+
+  const env = getEnv();
   const db = Knex(env);
   return db.migrate.latest(env.migrations)
-    .then(() => db.seed.run(env.seeds))
+    .then(() => {
+      const blockNumbers = sortIntegers(Object.keys(blockDetails));
```
epheph: `_(blockDetails).keys().map(Number).sortBy().value()`

Not sure if it's better than the function you've written though.

robert-e-davidson3: Yours is more concise so I'm going with it. Also TIL about `_(someNativeCollection)`!

PR #783, opened 2019-01-18
Don't start augur-node sync until the Ethereum node done syncing

ryanberckmans:
Fixes AugurProject/augur#260

Code is ready.

**But**, the Alchemy doesn't implement `eth_syncing`; they return status 400 "unsupported method", and ethrpc intercepts this 400 and emits a disconnected event which causes "Disconnected from Ethereum node", breaking everything.

The simplest fix here is to request that Alchemy implement eth_syncing, even if they just always return `false`. I'll request this of them and report back here.

### DIFF
--- a/src/blockchain/bulk-sync-augur-node-with-blockchain.ts
+++ b/src/blockchain/bulk-sync-augur-node-with-blockchain.ts
@@ -24,9 +24,10 @@ function setSyncFinished() {
   augurEmitter.emit(SubscriptionEventNames.SyncFinished);
 }
 
-export async function bulkSyncAugurNodeWithBlockchain(db: Knex, augur: Augur, blocksPerChunk: number|undefined): Promise<number> {
-  const row: HighestBlockNumberRow|null = await db("blocks").max("blockNumber as highestBlockNumber").first();
-  const lastSyncBlockNumber: number|null|undefined = row!.highestBlockNumber;
+export async function bulkSyncAugurNodeWithBlockchain(db: Knex, augur: Augur, blocksPerChunk: number | undefined): Promise<number> {
+  await waitForEthNodeToFinishSyncing(augur);
+  const row: HighestBlockNumberRow | null = await db("blocks").max("blockNumber as highestBlockNumber").first();
+  const lastSyncBlockNumber: number | null | undefined = row!.highestBlockNumber;
   const uploadBlockNumber: number = augur.contracts.uploadBlockNumbers[augur.rpc.getNetworkID()] || 0;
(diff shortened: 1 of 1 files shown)


src/blockchain/bulk-sync-augur-node-with-blockchain.ts
```diff
@@ -68,3 +75,29 @@ async function getHighestBlockNumber(augur: Augur): Promise<number> {
...
   });
 }
+
+async function getIsEthNodeSyncing(augur: Augur): Promise<{
+  highestBlockNumber: number;
+  isEthNodeSyncing: boolean;
+}> {
+  const getHighestBlockNumberPromise = getHighestBlockNumber(augur);
+  const syncingPromise = new Promise<any>((resolve, reject) => {
+    augur.rpc.eth.syncing(null, (error, response: any) => {
```
epheph: Can this be typed as boolean?

src/blockchain/bulk-sync-augur-node-with-blockchain.ts
```diff
@@ -24,9 +24,16 @@ function setSyncFinished() {
   augurEmitter.emit(SubscriptionEventNames.SyncFinished);
 }
 
-export async function bulkSyncAugurNodeWithBlockchain(db: Knex, augur: Augur, blocksPerChunk: number|undefined): Promise<number> {
-  const row: HighestBlockNumberRow|null = await db("blocks").max("blockNumber as highestBlockNumber").first();
-  const lastSyncBlockNumber: number|null|undefined = row!.highestBlockNumber;
+export async function bulkSyncAugurNodeWithBlockchain(db: Knex, augur: Augur, blocksPerChunk: number | undefined): Promise<number> {
+  while (true) {
```
epheph: Looks good. This might be a useful function to have broken out, I can imagine a few different places you might want to await for sync.

src/blockchain/bulk-sync-augur-node-with-blockchain.ts
ryanberckmans: No, because the result is an object if the node is syncing. https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing

My need to model the response type explicitly was limited by fact that this change only cares if its `=== false`.

src/blockchain/bulk-sync-augur-node-with-blockchain.ts
ryanberckmans: Great idea. Fixed next push

PR #784, opened 2019-01-18
Add 'yarn osx-copy-augur-app-db-to-augur-node-cli'

ryanberckmans:
This helper script will _overwrite_ your current augur-node cli database
with the DB from your OSX installed Augur App, and apply latest migrations
to the new cli db.

Why? In my workflow, my OSX installed Augur App is often at HEAD of
blockstream, but I frequently need to reset my augur-node CLI database.

### DIFF
--- a/knexfile.js
+++ b/knexfile.js
@@ -55,7 +55,7 @@ module.exports = {
   "prod-test": {
     client: "sqlite3",
     connection: {
-      filename: "./augur-1-2.db",
+      filename: "./augur-1-3.db",
     },
     migrations: {
       directory: "./src/migrations",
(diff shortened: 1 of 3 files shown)


epheph:
Seems like a handy tool. One thing I've run into with migrations is the whole .ts / .js thing. I run all my augur-node stuff with ts-node/register, so the migration files themselves are named .ts, but in the augur-app they are .js ? Do you run into this?

pgebheim:
This is definitely an issue that he brought up!

Wonder if there's some way to fix that??

ryanberckmans:
@pgebheim ready for merge

epheph:
```Yes! The script fixes that by updating the knex_migrations meta table.``` 😭 👍

PR #788, opened 2019-01-22
Serialize jest runs to see if it fixes sporadic runner timeouts.

epheph:
Fix coveralls while we are at it


epheph:
Increases time required from ~13s to ~39s

epheph:
I don't see this as a permanent solution, but one that should help see if there's some strange locking/concurrency thing in the runner (or sqlite?) occurring.

PR #789, opened 2019-01-22
add getters for user and system db versions

bthaile:
add getters for system db version and user's highest db version based on scanning user's data directory for `augur-1-`. In this case we are only interested in mainnet databases. This give information to augur-app to inform user when they have to resync.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -14,7 +14,7 @@ import { logger } from "./utils/logger";
 import { LoggerInterface } from "./utils/logger/logger";
 import { BlockAndLogsQueue } from "./blockchain/block-and-logs-queue";
 
-import { getFileHash } from "./sync/file-operations";
+import { getFileHash, getHighestDbVersions } from "./sync/file-operations";
 import { BackupRestore } from "./sync/backup-restore";
 import { checkOrphanedOrders } from "./blockchain/check-orphaned-orders";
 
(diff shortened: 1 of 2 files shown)


src/sync/file-operations.ts
```diff
@@ -71,3 +71,20 @@ export function fileCompatible(filename: string, networkId: string, dbVersion: n
   const compSyncfile = format(DB_WARP_SYNC_FILE_ENDING, networkId, dbVersion);
   return filename.endsWith(compSyncfile);
 }
+
+export function getHighestDbVersions(directoryDir: string, dbFileName: string): number {
```
pgebheim: Neato.

src/sync/file-operations.ts
```diff
@@ -72,7 +72,7 @@ export function fileCompatible(filename: string, networkId: string, dbVersion: n
   return filename.endsWith(compSyncfile);
 }
 
-export function getHighestDbVersions(directoryDir: string, dbFileName: string): number {
+export function getHighestDbVersion(directoryDir: string, dbFileName: string): number {
```
pgebheim: I didn't even see that ;-)

bthaile: initially I was going to return an array then the caller would decide what to do with it. but might as well just return highest like the name of method says.

src/controller.ts
```diff
@@ -139,6 +139,21 @@ export class AugurNodeController {
...
 
+  public systemDbVersion(): number {
+    return DB_VERSION;
+  }
+
+  public highestUserDbVersion(): number {
+    const version = 0;
+    try {
+      // only interested in mainnet databases
+      return getHighestDbVersion(this.databaseDir, "augur-1-");
```
epheph: Only works for mainnet?

bthaile: I don't we create if users have to resync testnets.

bthaile: updated to use augur networkId instead of hardcoding mainnet networkId.

PR #793, opened 2019-01-24
Fix short price unrealized realized

nuevoalex:
We want:

 - Avg price to always be the price of the long position even when for a short
 - Unrealized to produce the correct value using this avg price
 - Realized Profit to be recorded relative to the outcome being traded, regardless of long or short
 - Scalar profit to properly be calculated when shorting

This set of changes should cover that. Due to these changes we could remove the realized profit summing as well since that will be covered by the new realized profit behavior


bthaile:
@pg have you looked at this. I'd like a second opinion before merging this in.

PR #806, opened 2019-01-29
Deduct tokens from open orders

epheph:
Looks like an oversight. a creator can escrow shares or tokens, but only shares were being deducted as order-fills were coming in.


src/blockchain/log-processors/order-created.ts
```diff
@@ -49,6 +49,8 @@ export async function processOrderCreatedLog(augur: Augur, log: FormattedEventLo
       fullPrecisionPrice: fullPrecisionPrice.toString(),
       fullPrecisionAmount: fullPrecisionAmount.toString(),
       originalFullPrecisionAmount: fullPrecisionAmount.toString(),
+      originalTokensEscrowed: fixedPointToDecimal(moneyEscrowed, BN_WEI_PER_ETHER).toString(),
```
ryanberckmans: Small thing-- can we take the value `fixedPointToDecimal(moneyEscrowed, BN_WEI_PER_ETHER).toString()`, assign it to a variable, and use it to set both `originalTokensEscrowed` and `tokensEscrowed`?

epheph:
`Do we want originalTokensEscrowed and originalSharesEscrowed in the AN public API?` <- Sure we can do it right now. I really just wanted to make sure we start collecting it, but no reason not to add it to the getter now, since it doesn't change any value currently being used

src/blockchain/log-processors/order-created.ts
epheph: Sure thing

PR #807, opened 2019-01-29
Fix CompleteSets storage

epheph:
Before reading the following, please be at peace with the fact that everything, everything is terrible.

Terrible \# 1 - CompleteSets being stored in augur-node is completely non-functional right now
All rows for CompleteSets have NULL for `numPurchasedOrSold` and `numCompleteSets` because
Terrible \# 2 - passing in a BigNumber to the sql insert that expects a string results in a column set to null with no error or warning 
We didn't notice this because the UI doesn't use getters for CompleteSets and...
Terrible \# 3 - The test itself noticed the problem, but was not awaiting appropriately. the `expect()` was passed a promise, but the result of the expect itself was not awaited, so the test was passing before the tested code actually ran. We didn't notice because:
Terrible \# 4 - Unhandled exceptions are simply printed to console.warn without impacting overall test run's exit status and
Terrible \# 6 - We don't inspect unhandled promises because we only do that in tslint and not the eslint covering JS

So the fix was simple, just toString() the BN and await the promises. But then the question is why did we allow a null in DB anyway, where it surely should always be a number. So i tried to add a migration to change the column, but then:
Terrible \# 7 - SQLite doesn't support changing column types, even just adding `NOT NULL`

The proposed SQLite solution to this is to dump, recreate with new schema, then reload the table and swap it into place. Doesn't seem worth it for type-safety of a field we don't yet use, so I modified the existing migration for new deployments.


pgebheim:
Wow that is just a pile of terrible right there. We've had a number of problems stemming from BN -> Knex being inserted as NULL.

iirc this is because BN has toValue and Knex tries to call that....

PR #812, opened 2019-02-04
Fixes augur#811

pgebheim:
We didn't manually convert from hex to int during reorg, so the code works on sqlite but not postgres.

### DIFF
--- a/src/blockchain/process-block.ts
+++ b/src/blockchain/process-block.ts
@@ -64,7 +64,7 @@ export async function processBlockAndLogs(db: Knex, augur: Augur, direction: Blo
       await dbWritesFunction(trx);
       await db("transactionHashes")
         .transacting(trx)
-        .where({ blockNumber: block.number })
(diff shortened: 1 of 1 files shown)


pgebheim:
@FilipPyrek  -- Mind testing out this patch?

FilipPyrek:
@pgebheim yep, gonna test it. I'will let you know tomorrow (I'm based in Europe..) 👍

src/blockchain/process-block.ts
```diff
@@ -64,7 +64,7 @@ export async function processBlockAndLogs(db: Knex, augur: Augur, direction: Blo
       await dbWritesFunction(trx);
       await db("transactionHashes")
         .transacting(trx)
-        .where({ blockNumber: block.number })
+        .where({ blockNumber: parseInt(block.number, 16) })
```
epheph: You're right, sqlite transparently converts hex-format to decimal!

PR #817, opened 2019-02-12
Fix migration to be compatible with Postgres

FilipPyrek:
Another pull request from me fixing migrations to be compatible with Postgres 🙂


FilipPyrek:
Hi, what's the status here?

pgebheim:
Lgtm

pgebheim:
Just a double check from @nuevoalex and then we can merge!

Thanks again @FilipPyrek

PR #821, opened 2019-02-24
frozen funds

ryanberckmans:
https://github.com/AugurProject/augur/issues/988

(frozen funds algorithms in this PR are updated from those mentioned in the ticket)


src/server/getters/get-profit-loss.ts
```diff
@@ -384,6 +388,9 @@ export async function getProfitLossSummary(db: Knex, augur: Augur, params: GetPr
       averagePrice: startProfit.averagePrice,
       outcome: startProfit.outcome,
       marketId: startProfit.marketId,
+      // TODO Alex - are these right?
+      frozenFunds: startProfit.frozenFunds,
```
nuevoalex: This should be negated

src/blockchain/log-processors/profit-loss/frozen-funds.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (params.event === "ClaimProceeds") {
+    return {
+      // TODO doc
+      frozenFunds: ZERO,
+      frozenProfit: ZERO,
+    };
+  }
+  // params.event is a trade executed; TODO break each event into a separate function
+  const trade = params.event.tradeData;
+  const frozenFundsAfterEvent = Object.assign({}, params.frozenFundsBeforeEvent);
```
nuevoalex: why an object? seems like we only use the `frozenFunds` attribute so why not just have `frozenFunds`?

src/blockchain/log-processors/profit-loss/frozen-funds.ts
```diff
@@ -0,0 +1,99 @@
...
+      creatorTokens: ZERO,
+      fillerTokens: ZERO,
+    };
+  }
+  */
+  const tokensFreedFromEscrowByThisTrade = BigNumber.min(trade.numCreatorShares, trade.numFillerShares).multipliedBy(ZERO /* TODO displayRange */);
+
+  // TODO Alex - can you confirm formula for this:
+  const creatorPortionOfTokensFreedFromEscrow = tokensFreedFromEscrowByThisTrade; // TODO actual value I think is f(minPrice, maxPrice, price)?
+  const fillerPortionOfTokensFreedFromEscrow = tokensFreedFromEscrowByThisTrade; // TODO actual value I think is f(minPrice, maxPrice, price)?
```
nuevoalex: Whoever was long paid the exact price of the trade so the way Id structure this is

`
numCompleteSetsSold = BigNumber.min(trade.numCreatorShares, trade.numFillerShares);
longPay = price * numCompleteSetsSold;
shortPay = tokensFreedFromEscrowByThisTrade - longPay;
`

Where filler and creator portions are assigned based on who was long and who was short

src/blockchain/log-processors/profit-loss/frozen-funds.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  if (frozenFundsAfterEvent.frozenFunds.isGreaterThanOrEqualTo(
+    params.frozenFundsBeforeEvent.frozenFunds)) {
+    // frozen funds increased or stayed the same, so my available funds decreased or
+    // stayed the same, which means that my frozen profit should increase by realized
+    // profit and loss for this trade. Ie. frozen profit is profit that I could claim,
+    // but haven't, and we detect this by looking for a non-increase in available funds.
+    // TODO Alex - is this right?
+    frozenFundsAfterEvent.frozenProfit = frozenFundsAfterEvent.frozenProfit.plus(params.event.realizedProfit);
```
nuevoalex: frozen profit doesn't need to be stored outside of frozen funds, its just a part of calculating the frozen funds and wont be displayed or used in its own right. This should just be `frozenFunds += params.event.realizedProfit`

frozen profit is profit or loss which is not available in the users funds as it only applied in the context of adjusting the entry into a new position. It's not profit that you could claim but haven't, that's _unrealized_ profit.

src/blockchain/log-processors/profit-loss/update-profit-loss.ts
```diff
@@ -98,6 +104,15 @@ export async function updateProfitLoss(db: Knex, marketId: Address, positionDelt
     newPrice = (oldPrice.multipliedBy(oldPosition.abs())).plus(price.multipliedBy(positionDelta.abs())).dividedBy(position.abs());
   }
 
+  const nextFrozenFunds = getFrozenFundsAfterEventForOneOutcome({
+    frozenFundsBeforeEvent: {
+      frozenFunds: oldFrozenFunds,
+      frozenProfit: oldFrozenProfit,
+    },
+    // TODO Alex - is this right profit from "frozenProfit += available_funds decreased ? realized PL : 0"?
```
nuevoalex: This isn't right. frozen profit shouldn't be stored or incremented over time. its just part of calculating frozen funds

src/migrations/20190201155650_wcl_profit_loss_timeseries.ts
```diff
@@ -26,18 +30,36 @@ exports.up = async (knex: Knex): Promise<any> => {
     table.specificType("price", "varchar(255) NOT NULL CONSTRAINT nonnegativeAmount CHECK (ltrim(\"price\", '-') = \"price\")");
     table.string("position", 42).notNullable();
     table.string("profit", 255).defaultTo("0");
+    table.string("frozenFunds", 255).defaultTo("0");
+    table.string("frozenProfit", 255).defaultTo("0");
```
nuevoalex: Frozen profit should not be stored in the DB

src/blockchain/log-processors/profit-loss/frozen-funds.ts
```diff
@@ -0,0 +1,99 @@
...
+  // TODO Alex - can you confirm formula for this:
+  const creatorPortionOfTokensFreedFromEscrow = tokensFreedFromEscrowByThisTrade; // TODO actual value I think is f(minPrice, maxPrice, price)?
+  const fillerPortionOfTokensFreedFromEscrow = tokensFreedFromEscrowByThisTrade; // TODO actual value I think is f(minPrice, maxPrice, price)?
+
+  if (params.event.creatorOrFiller === "creator") {
+    // I created this trade
+    // TODO doc formulas
+    frozenFundsAfterEvent.frozenFunds = frozenFundsAfterEvent.frozenFunds.minus(creatorPortionOfTokensFreedFromEscrow);
+    frozenFundsAfterEvent.frozenFunds = frozenFundsAfterEvent.frozenFunds.plus(trade.numCreatorTokens);
+    frozenFundsAfterEvent.frozenFunds = frozenFundsAfterEvent.frozenFunds.minus(trade.numFillerTokens);
```
nuevoalex: Realizing now this won't quite work since the goal is to figure out how many tokens the traders actually received and this won't always coincide with that.

I think the best way may be to actually look at shares provided and the price.

So For tokens spent you can still use `frozenFunds = .frozenFunds.plus(trade.numCreatorTokens);`

but then we can avoid the complete set calc entirely and the above line and instead just do something like:

`tokensReceived = trade.numCreatorShares.multipliedBy(price);
frozenFunds = frozenFunds.minus(tokensReceived);`

where price will be `maxPrice - price` for the party shorting (whichever that is)

ryanberckmans:
@nuevoalex ready for review. Per our chat I'll next add test cases from chwy's spreadsheet

src/blockchain/log-processors/profit-loss/frozen-funds.ts
```diff
@@ -0,0 +1,87 @@
...
+  const myTokensReceived = mySharesSent.multipliedBy(tokensReceivedPrice);
+  const myTokensSent = trade.creatorOrFiller === "creator" ? trade.numCreatorTokens : trade.numFillerTokens;
+
+  // Tokens received are subtracted from frozen funds because the
+  // user now possesses those funds, and thus those funds are no longer
+  // "frozen". Similarly, tokens sent are added to frozen funds.
+  frozenFundsAfterEvent = frozenFundsAfterEvent.minus(myTokensReceived);
+  frozenFundsAfterEvent = frozenFundsAfterEvent.plus(myTokensSent);
+
+  if (frozenFundsAfterEvent.isGreaterThanOrEqualTo(
```
nuevoalex: As I mentioned I suspect we should actually remove this condition and just apply profit any time there if a non-zero PL realized. Tests should reveal this is correct once they've been added

ryanberckmans:
Added all test cases from chwy's WCL sheet to integration and unit tests. Fixed two formula bugs. This is nearly ready to merge, just need to ask chwy about Cat3-Tr3 State7.

PR #822, opened 2019-02-27
Temporary DB must be explicitly destroyed

justinbarry:


justinbarry:
NVM. Going a different direction.

PR #823, opened 2019-02-27
Destroy temporary DB when finished

morninglite:
When creating warp sync files, a temporary copy of the DB is made. The Knex object associated with this temporary DB keeps it open, even after the file is deleted, so the space used is never freed by the operating system. Since this happens on a regular basis (every 100 blocks), eventually either all available space will be consumed or the augur process will run out of file descriptors.

This change seems to fix that.


justinbarry:
@morninglite Nice find!

bthaile:
We’ll update and merge

PR #830, opened 2019-03-14
augur-1192 ads platform activity stats to api

brianosaurus:
#### Ads platform activity stats to the API

Provides
-Active Users
- number of Trades
- Open Interest
- Markets Created
- Volume
- Money at Stake


test/unit/server/getters/get-platform-activity-stats.js
```diff
@@ -0,0 +1,33 @@
+const { setupTestDb, seedDb } = require("test.database");
```
ryanberckmans: Mind if we do tests in TypeScript, too?

brianosaurus: Wow. I thought jest didn't support that but it appears I'm wrong.

I can migrate the tests to ts on the followup PRs that will have the knex stuff sorted out.

PR #831, opened 2019-03-18
Add P&L percents and financial math library

ryanberckmans:
I'll write a better commit msg for the merge.

We'd like to merge this now so UI team can continue testing. After the merge I'll continue working on this:

1. clean up comments in financial-math and dimension-quantity
2. move financial-math to its own npm module or monorepo subdir ?? (might name it `augur-math`)
3. move dimension-quantity to its own npm module (might name it `ts-quantity`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "replace-in-file": "3.0.0",
     "tslint": "5.7.0",
     "tslint-eslint-rules": "4.1.1",
-    "typescript": "3.1.6"
+    "typescript": "3.3.3333"
   },
   "buildDependencies": [
     "typescript",
--- a/src/server/getters/get-profit-loss.ts
+++ b/src/server/getters/get-profit-loss.ts
@@ -9,7 +9,7 @@ import { getCurrentTime } from "../../blockchain/process-block";
 import { ZERO } from "../../constants";
 import { Address } from "../../types";
 import { Percent, Price, Shares, Tokens } from "../../utils/dimension-quantity";
-import { positionGetCurrentValue, positionGetRealizedProfitPercent, positionGetTotalProfit, positionGetTotalProfitPercent, positionGetUnrealizedProfit, positionGetUnrealizedProfitPercent, positionGetUnrealizedCost } from "../../utils/financial-math";
+import { positionGetCurrentValue, positionGetRealizedProfitPercent, positionGetTotalCost, positionGetTotalProfit, positionGetTotalProfitPercent, positionGetUnrealizedCost, positionGetUnrealizedProfit, positionGetUnrealizedProfitPercent } from "../../utils/financial-math";
 
 const DEFAULT_NUMBER_OF_BUCKETS = 30;
 
@@ -84,6 +84,7 @@ export interface ProfitLossResult extends
   total: BigNumber; // total profit in tokens (eg. ETH). Always equal to realized + unrealized
   unrealizedCost: BigNumber;
   realizedCost: BigNumber;
+  totalCost: BigNumber;
   realizedPercent: BigNumber; // TODO from 0 to 1, not 0 to 100
   unrealizedPercent: BigNumber; // TODO from 0 to 1, not 0 to 100
   totalPercent: BigNumber; // TODO from 0 to 1, not 0 to 100
@@ -216,7 +217,7 @@ function getProfitResultsForTimestamp(plsAtTimestamp: Array<ProfitLossTimeseries
     const realizedCost = new Tokens(outcomePl.realizedCost);
     const realizedProfit = new Tokens(outcomePl.profit);
     const outcome = outcomePl.outcome;
-    const averagePrice = new Price(outcomePl.price);
+    const averagePerSharePriceToOpenPosition = new Price(outcomePl.price);
     const minPrice = new Price(outcomePl.minPrice);
     const netPosition = new Shares(outcomePl.position);
     const frozenFunds = new Tokens(outcomePl.frozenFunds);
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -123,7 +123,7 @@
     "replace-in-file": "3.0.0",
     "tslint": "5.7.0",
     "tslint-eslint-rules": "4.1.1",
-    "typescript": "3.1.6"
+    "typescript": "3.3.3333"
```
brianosaurus: Somebody has a sense of humor.

src/server/getters/get-profit-loss.ts
```diff
@@ -46,11 +49,12 @@ export interface ProfitLossTimeseries extends Timestamped, FrozenFunds {
   marketId: Address;
   outcome: number;
   transactionHash: string;
-  price: BigNumber;
-  position: BigNumber;
+  price: BigNumber; // denominated in tokens/share. average price user paid for shares in the current open position
+  position: BigNumber; // denominated in shares. Known as "net position", this is the number of shares the user currently owns for this outcome; if it's a positive number, the user is "long" and earns money if the share price goes up; if it's a negative number the user is "short" and earns money if the share price goes down. Eg. "-15" means an open position of short 15 shares.
```
brianosaurus: Nice comment...helpful.

src/server/getters/get-profit-loss.ts
```diff
@@ -60,14 +64,29 @@ export interface OutcomeValueTimeseries extends Timestamped {
...
+  averagePrice: BigNumber; // denominated in tokens/share. average price user paid for shares in the current open position
+  realized: BigNumber; // realized profit in tokens (eg. ETH) user already got from this market outcome. "realized" means the user bought/sold shares in such a way that the profit is already in the user's wallet
+  unrealized: BigNumber; // unrealized profit in tokens (eg. ETH) user could get from this market outcome. "unrealized" means the profit isn't in the user's wallet yet; the user could close the position to "realize" the profit, but instead is holding onto the shares. Computed using last trade price.
+  total: BigNumber; // total profit in tokens (eg. ETH). Always equal to realized + unrealized
+  unrealizedCost: BigNumber; // denominated in tokens. Cost of shares in netPosition
+  realizedCost: BigNumber; // denominated in tokens. Cumulative cost of shares included in realized profit
+  totalCost: BigNumber; // denominated in tokens. Always equal to unrealizedCost + realizedCost
+  realizedPercent: BigNumber; // realized profit percent (ie. profit/cost)
```
brianosaurus: Can these be calculated? Have them be functions instead of data on an interface?

src/server/getters/get-profit-loss.ts
```diff
@@ -190,36 +232,85 @@ function bucketAtTimestamps<T extends Timestamped>(timestampeds: Dictionary<Arra
...
+    const frozenFunds = new Tokens(outcomePl.frozenFunds);
+
+    // averagePrice is average price per share user paid to open this position;
+    // but, we want to display averagePrice; for scalar markets this means adding
+    // minPrice. Eg. if a user paid an average price of 4.5 for a scalar market, but
+    // the minPrice in that market is 3, we actually need to show 4.5+3=7.5 in the UI.
+    // TODO use new tradeDisplayPrice() from financialMath?
+    const averagePriceDisplay: Price = averagePricePerShareToOpenPosition.plus(minPrice);
+
+    const lastPrice: Price | undefined = outcomeValuesAtTimestamp ? new Price(outcomeValuesAtTimestamp[outcome].value).minus(minPrice) : undefined;
```
brianosaurus: Why undefined instead of null? Just curious.

src/server/getters/get-user-trading-positions.ts
```diff
@@ -169,3 +187,51 @@ async function getEthEscrowedInValidityBonds(db: Knex, marketCreator: Address):
   }
   return totalValidityBonds;
```
brianosaurus: It looks like this method returns Promise<BigNumber> but ZERO is defined as new BigNumber(0) which isn't a promise. Also, I am unsure if marketRow has data where totalValidityBonds turns into a promise? Maybe I'm missing something.

src/utils/dimension-quantity.ts
```diff
@@ -0,0 +1,316 @@
...
+// **** specific units below here; a lot of this could be generated code
+
+export function scalar(magnitude: number | BigNumber): Scalar {
+  if (typeof magnitude === "number") {
+    return new Scalar(new BigNumber(magnitude));
+  }
+  return new Scalar(magnitude);
+}
+const ScalarUnitDimension: DimensionVector = Object.freeze({
+  tokens: 0,
```
brianosaurus: Why not the constant ZERO here? Just curious.

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
+    "noFallthroughCasesInSwitch": true,
```
brianosaurus: This might cause issues elsewhere because I recall seeing a fall through being used somewhere. Could be wrong though...It certainly might catch someone off guard? Dunno...

tslint.json
```diff
@@ -105,6 +105,9 @@
     "no-shadowed-variable": false,
     "object-literal-sort-keys": false,
     "array-type": [true, "generic"],
+    "no-consecutive-blank-lines": false,
+    "no-trailing-whitespace": false,
```
brianosaurus: Hmm....

src/server/getters/get-profit-loss.ts
ryanberckmans: Yes! Agree we should JIT everything where possible. In this case `ProfitLossResult` is the base type for the (serialized) RPC response type for getUserTradingPositions() et al

src/server/getters/get-profit-loss.ts
ryanberckmans: `undefined` seems more technically correct in my head. I pretty much never use null.

src/server/getters/get-user-trading-positions.ts
ryanberckmans: Async functions have syntax sugar to wrap returned values in `Promise.resolve`.

From recent PR

```
  return new Promise<PlatformActivityResult>((resolve, reject) => {
    resolve(result);
  });
```

In any function (including non-async) this could instead be:

```
return Promise.resolve(result);
```

Or in an async function:

```
return result; // automagically wrapped in Promise.resolve()
```

src/utils/dimension-quantity.ts
ryanberckmans: `ZERO` is a BigNumber. Opted to use regular javascript `number` for the dimension exponents. During implementation was experimenting with including exponent information some types, using typescript "literal types" which require numbers. Am currently satisfied to use regular numbers for exponents; they exception/debug print easier and require fewer allocations, and there's no risk of overflow or need for fancy operations for exponents.

tsconfig.json
ryanberckmans: afiact a violation of `noFallthroughCasesInSwitch` will result in build breaking and not a silent error.

tslint.json
ryanberckmans: Found these whitespace rules to be a big distraction in my editor. I wish there was a setting to "auto fix but never warn", like `gofmt`. afaik there's no such setting-- tslint will only warn AND fix... the warnings are an attention drain, prefer to disable them.

PR #854, opened 2019-04-22
Add blockNumber, logIndex to DB table outcome_value_timeseries

ryanberckmans:
The DB table outcome_value_timeseries uses timestamp as the sort column, but
trades in the same block may have the same timestamp, preventing clients of
outcome_value_timeseries from determining the correct trade order.

Also sort by (timestamp, blockNumber, logIndex) instead of just timestamp in
queryOutcomeValueTimeseries, the only client of outcome_value_timeseries.

Fixes AugurProject/augur#1634


bthaile:
Not using logIndex has gotten me in the past. 

Iiuc, augur-node currently throws away logIndex?
And you are adding it, how would the user get logIndex on existing augur-node dbs? Would they have to fully resync?

PR #857, opened 2019-04-27
Restore uniqueless by (market, outcome, timestamp) in queryOutcomeValueTimeseries

ryanberckmans:
Prior to cae512a4074ae7777ddac7a8b31125fb75edf5d0, queryOutcomeValueTimeseries
had the postcondition that Array<OutcomeValueTimeseries> was unique in
(market, outcome, timestamp).

This change restores that postcondition while ensuring that the most recent of
each (market, outcome, timestamp) is included in the result.

Related to AugurProject/augur#1849
Follow-up to cae512a4074ae7777ddac7a8b31125fb75edf5d0

### DIFF
--- a/src/server/getters/get-profit-loss.ts
+++ b/src/server/getters/get-profit-loss.ts
@@ -211,20 +211,39 @@ async function queryProfitLossTimeseries(db: Knex, now: number, params: GetProfi
   return await query;
 }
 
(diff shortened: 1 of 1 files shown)


ryanberckmans:
This fix works but the problem it fixes hasn't occurred in practice, closing to KISS

PR #860, opened 2019-04-30
Handle open orders and self-trades in frozen funds

ryanberckmans:
Frozen funds now includes tokens escrowed for open orders.

Frozen funds now handles self-filled trades: if a trade is self-filled (ie. user
traded with themselves), then the OrderCreated log looks the same as if the
trade was not self-filled, but the contracts silently skip creating complete
sets, instead returning tokens to the user. So augur-node deducts these tokens
from frozen funds to handle this on-chain exception.

AugurProject/augur#1869 AugurProject/augur#1852

### DIFF
--- a/src/blockchain/log-processors/order-canceled.ts
+++ b/src/blockchain/log-processors/order-canceled.ts
@@ -1,15 +1,18 @@
 import Augur from "augur.js";
-import * as Knex from "knex";
 import { BigNumber } from "bignumber.js";
-import { Bytes32, FormattedEventLog, OrderState } from "../../types";
-import { augurEmitter } from "../../events";
+import * as Knex from "knex";
 import { SubscriptionEventNames } from "../../constants";
+import { augurEmitter } from "../../events";
+import { getDefaultPLTimeseries, ProfitLossTimeseriesRow } from "../../server/getters/get-profit-loss";
+import { Bytes32, FormattedEventLog, OrdersRow, OrderState } from "../../types";
+import { FrozenFunds, getFrozenFundsAfterEventForOneOutcome } from "./profit-loss/frozen-funds";
+import { getCurrentTime } from "../process-block";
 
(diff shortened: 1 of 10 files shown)


src/blockchain/log-processors/order-canceled.ts
```diff
@@ -20,17 +23,65 @@ interface MarketNumOutcomes {
...
+    if (!prevProfitLoss) throw new Error(`expected to find previous wcl_profit_loss_timeseries on order cancel orderId=${log.orderId}`);
+
     const orderTypeLabel = log.orderType === "0" ? "buy" : "sell";
     await db.from("orders").where("orderId", log.orderId).update({ orderState: OrderState.CANCELED });
-    await  db.into("orders_canceled").insert({ orderId: log.orderId, transactionHash: log.transactionHash, logIndex: log.logIndex, blockNumber: log.blockNumber });
-    const ordersRow: MarketIDAndOutcomeAndPrice = await  db.first("marketId", "outcome", "price", "sharesEscrowed", "orderCreator").from("orders").where("orderId", log.orderId);
+    await db.into("orders_canceled").insert({ orderId: log.orderId, transactionHash: log.transactionHash, logIndex: log.logIndex, blockNumber: log.blockNumber });
+
+    const nextFrozenFunds: FrozenFunds = getFrozenFundsAfterEventForOneOutcome({
+      frozenFundsBeforeEvent: prevProfitLoss,
```
nuevoalex: Should `frozenFundsBeforeEvent` be labeled `profitLossBeforeEvent` maybe?

ryanberckmans: I think maybe it is a different thing, `frozenFundsBeforeEvent` is a member of `FrozenFundsParams` and `prevProfitLoss` is being downcast into that type

ryanberckmans:
Closing PR to fix bug with self-filled trades + frozen funds

PR #861, opened 2019-04-30
Handle open orders and self-trades in frozen funds

ryanberckmans:
Frozen funds now includes tokens escrowed for open orders.

Frozen funds now handles self-filled trades: if a trade is self-filled (ie. user
traded with themselves), then the OrderCreated log looks the same as if the
trade was not self-filled, but the contracts silently skip creating complete
sets, instead returning tokens to the user. So augur-node deducts these tokens
from frozen funds to handle this on-chain exception.

AugurProject/augur#1869 AugurProject/augur#1852


ryanberckmans:
Alex approved this in the other closed PR ( https://github.com/AugurProject/augur-node/pull/860 )

PR #871, opened 2019-05-15
tradeGroupId should be a string

bthaile:
Getting errors in the migration script in augur app.

image

it is only there to match up orders not calculations are performed on this field. tradeGroupId is a UUID


bthaile:
closer look, tradeGroupId is null and the constraint doesn't allow nulls. closing PR

PR #876, opened 2019-05-16
Add experimental invalid filter

ryanberckmans:
Experimental invalid filter which is the idea that although v1 cannot measure
market invalidity we can still detect an order book that will profit if the
market resolves invalid.

Experimental invalid filter is implemented as new DB markets.invalidROIPercent
and outcomes.invalidROIPercent.

outcomes.invalidROIPercent is the return on investment percent for that
outcome's best bid/ask if the market resolves invalid.

markets.invalidROIPercent is the max of its outcomes' invalidROIPercent.

The client API is a new get-markets filter parameter
`params.maxInvalidROIPercent`.  A recommended starter value is
`maxInvalidROIPercent=0.001`.

Related to AugurProject/augur#2155


ryanberckmans:
This is ready to merge when Phoebe is done testing

PR #894, opened 2019-06-29
Liquidity fixes

ryanberckmans:
Rinkeby sync works.

Gas price fetching turned back on with 1500ms timeout.

In rare cases Rinkeby DB migration may take up to **ten minutes**.
* because the liquidity metrics are updated via DB migration each time the algorithm changes and some DBs may do this up to a half dozen times (redundantly) depending on the DB's initial state

This release, Augur App customers will have a mainnet DB migration 30 to 90 seconds, similar to last release. Augur App shows "Connecting to Ethereum..." during this process


src/controller.ts
```diff
@@ -48,7 +48,7 @@ export class AugurNodeController {
   }
 
   public async start(errorCallback: ErrorCallback | undefined) {
-    // startFetchingGasPrice(); removing b/c of runtime issue in AWS
+    startFetchingGasPrice();
```
bthaile: Move this call to after block stream processing starts, no need to get gas prices during bulk sync because the websocket server isnt even started so augur-node wont be accepting requests.

bthaile:
Also explain why is the invalidity liquidity calculations occurring during bulk block processing? Why not do all the calc and db stuff after bulk blocking processing and during block stream processing? Don’t these calculations just need to be against the most current order book?

petong:
I just finished a sync from scratch, and while it looked like it had completed:

```Need to check 3 orderbooks for orphans
Need to check 2 orderbooks for orphans
Need to check 1 orderbooks for orphans
Need to check 0 orderbooks for orphans
Bulk orphaned orders check with blockchain complete.
Starting blockstream at block  4657905
new block: 4657905, 1561998382 (Mon Jul 01 2019 09:26:22 GMT-0700 (Pacific Daylight Time))```

Our status check says the service is not healthy:

```curl -v localhost:9001/status/blockage
* Connection #0 to host localhost left intact
{"status":"down","timestampDelta":11009,"blockNumber":4657905,"blockHash":"0x52084fa150a529c4e02a64cef86ae26937f644c0206ae0a6b12e9352f15b6764","timestamp":1561998382,"bulkSync":0}*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 9001 (#0)
> GET /status/blockage HTTP/1.1
> Host: localhost:9001
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 503 Service Unavailable
< X-DNS-Prefetch-Control: off
< X-Frame-Options: SAMEORIGIN
< X-Download-Options: noopen
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< Content-Type: application/json; charset=utf-8
< Content-Length: 179
< ETag: W/"b3-A/4Db8TBM4DzRGkdLlpuf/0HTUk"
< Date: Mon, 01 Jul 2019 19:29:51 GMT
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
{"status":"down","timestampDelta":11010,"blockNumber":4657905,"blockHash":"0x52084fa150a529c4e02a64cef86ae26937f644c0206ae0a6b12e9352f15b6764","timestamp":1561998382,"bulkSync":0}```

at the time the sync finished, last block on rinkeby was `4658628`


it never moves past that last block. stuck.

petong:
These changes worked for me locally, LGTM!

PR #900, opened 2019-09-04
minimum initial rep stake filter

nuevoalex:
add minimum initial REP stake filter to getMarkets


src/server/getters/get-markets.ts
```diff
@@ -78,6 +79,15 @@ export async function getMarkets(db: Knex, augur: {}, params: t.TypeOf<typeof Ge
     query.whereRaw("markets.endTime < ?", [params.maxEndTime]);
   }
 
+  if (params.minInitialRep) {
```
bthaile: I thought this was going to be a default filter and not changes were needed to the UI

nuevoalex: I think that would probably be better personally. The UI has to support a filter anyway in case people want to ignore this at a minimum, however it also makes more sense to me to apply the business logic rules we're applying in the UI like we do with other filters.

bthaile: I was under the impression there wouldn't be any UI changes b/c augur-node would handle the algo to filter out markets. I'm fine either way.

If we are going to allow the user to turn on|off this filter then we'll need a little bit of design work and UI work.

bthaile: Can you add the IR balance on the MarketsInfo object returned by getMarketsInfo getter? We can show it in the UI.

nuevoalex: Sure!

bthaile: also add the IR address, that would be all the data the UI needs to show.

PR #924, opened 2019-12-26
bump cut off time to Jan 15

bthaile:
bump cut off time to Jan 15, this will reduce the amount of REP needed to stake and markets pass filter, in the short term.


bthaile:
closing this, cut off was updated on a diff PR

